PR #1, opened 2022-09-24
Java client implementation

coder966:
This is a Java client implementation, which provide sync & async APIs.


tobiemh:
Awesome @coder966 I'm just going to make some comments - then we can get this merged!

src/main/java/com/surrealdb/connection/SurrealWebSocketConnection.java
```diff
@@ -0,0 +1,148 @@
...
+    private final Gson gson;
+    private final Map<String, CompletableFuture<?>> callbacks;
+    private final Map<String, Type> resultTypes;
+
+    // precomputed private variables
+    private final Pattern RECORD_ALREADY_EXITS_PATTERN = Pattern.compile("There was a problem with the database: Database record `(.+):(.+)` already exists");
+
+    @SneakyThrows
+    public SurrealWebSocketConnection(String host, int port){
+        super(URI.create("ws://"+host+":"+port+"/rpc"));
```
tobiemh, line 39: This might be `wss://` in production environments, so it might be better to allow passing the full url?

src/main/java/com/surrealdb/connection/model/RpcRequest.java
```diff
@@ -0,0 +1,21 @@
...
+ * @author Khalid Alharisi
+ */
+@Getter
+public class RpcRequest {
+    private final String id;
+    private final String method;
+    private final Object[] params;
+
+    public RpcRequest(String method, Object... params) {
+        id = UUID.randomUUID().toString(); // TODO: change to a faster random ID
```
tobiemh, line 17: This doesn't need to be cryptographically secure. The Node.js and Python libraries just use an incrementing number for each request. This is only necessary to detect which response is for which request.

src/main/java/com/surrealdb/driver/model/patch/CopyPatch.java
```diff
@@ -0,0 +1,19 @@
+package com.surrealdb.driver.model.patch;
```
tobiemh, line 1: We don't support the `Copy` JSON Patch type in SurrealDB.

src/main/java/com/surrealdb/driver/model/patch/MovePatch.java
```diff
@@ -0,0 +1,19 @@
+package com.surrealdb.driver.model.patch;
```
tobiemh, line 1: We don't support the `Move` JSON Patch type in SurrealDB.

src/main/java/com/surrealdb/driver/model/patch/TestPatch.java
```diff
@@ -0,0 +1,19 @@
+package com.surrealdb.driver.model.patch;
```
tobiemh, line 1: We don't support the `Test` JSON Patch type in SurrealDB.

src/main/java/com/surrealdb/driver/model/patch/ReplacePatch.java
```diff
@@ -0,0 +1,19 @@
+package com.surrealdb.driver.model.patch;
```
tobiemh, line 1: We actually have a custom Patch type `change` which is used for diff-match-patching of text values in a record. It's basically the same as the `replace` Patch, but the `value` is a diff-match-patch string.

tobiemh (changes requested):
This is a good starting point. There will definitely be changes around authentication/signup/signin, but those can come after the documentation improvements!

src/main/java/com/surrealdb/driver/model/patch/ReplacePatch.java
tobiemh: Just to be clear, this is IN ADDITION to the `replace` patch! So SurrealDB would have:

1. add
2. remove
3. replace
4. change

src/main/java/com/surrealdb/connection/SurrealWebSocketConnection.java
coder966: I added a boolean parameter `useTls` that controls the protocol used. I think these params (hostname, port, useTLS?) are simpler for the user since the user might use an unsupported protocol like `http` or forget to append the `/rpc` path.

src/main/java/com/surrealdb/driver/model/patch/ReplacePatch.java
coder966: Great, I updated the PR with the changes requested.

tobiemh (approved):

coder966:
> This is a good starting point. There will definitely be changes around authentication/signup/signin, but those can come after the documentation improvements!

You are right, the authentication apis are incomplet. I couldn't find documentation on how the singup/authentication really works. I am looking forward to the docs.

Also NS, DB. Are they created on the fly? Live queries didn't work for me as well. But this is another topic which I have posted a question about in the discussion community https://github.com/surrealdb/surrealdb/discussions/1208.

tobiemh:
We'll sort out the authentication in future changes/PRs when the docs are improved @coder966! This is an awesome start 😀👏!

PR #3, opened 2022-10-06
Github Actions: Continuous Integration & Version Release

coder966:
In this PR:
- Add workflow to run tests on pushes/PRs to `main`.
- Add workflow to run tests and create a release when tags are created with name following the pattern `v*.*.*`.
- prettier log report of tests


.github/workflows/ci.yml
```diff
@@ -1,14 +1,23 @@
-name: Build
+name: Continuous Integration
```
phughk, line 1: Based on the other repositories I think this should be just `test`.
https://github.com/surrealdb/surrealdb.js/blob/main/.github/workflows/test.yml#L1
https://github.com/surrealdb/surrealdb.go/blob/main/.github/workflows/test.yml#L1

.github/workflows/ci.yml
```diff
@@ -1,14 +1,23 @@
...
 
 on:
   workflow_dispatch:
   push:
-    branches: [ master ]
+    branches:
+      - main
+  pull_request:
```
phughk, line 8: We actually don't want pull requests built I think. At least for the time being. The reason is that people could write bots that maliciously open a lot of PRs. So I would opt for having `pull_requests:` without any nested values, just like in the other repos.

.github/workflows/ci.yml
```diff
@@ -1,14 +1,23 @@
...
 
 jobs:
-  build:
+
+  test:
     runs-on: ubuntu-latest
     steps:
+
+      - name: Setup SurrealDB
+        run: docker run -d --rm --network host surrealdb/surrealdb:latest start --user=root --pass=root
```
phughk, line 19: You need to expose ports here, like in this repo
https://github.com/surrealdb/surrealdb.go/blob/main/.github/workflows/test.yml#L28

.github/workflows/ci.yml
```diff
@@ -18,13 +27,7 @@ jobs:
           java-version: '8'
           distribution: 'temurin'
 
-      - name: Build a Fat JAR
```
phughk, line 21: We still want the fat jar step. Otherwise we aren't uploading an artifact that includes all the necessary dependencies. We can run the tests before building the fat jar as I don't the tests depend on the fat jar.

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
+name: Version Release
```
phughk, line 1: Ah I see what you are doing. Thanks for writing this, it's really cool! I hope it wasn't taking too much of your time. We really appreciate community contributions ❤️ and I can see you put a lot of effort into this for the community. The way releases work in github is not by having a specific release pipeline (like you would in other build services), but instead by pointing at a specific artifact and giving it a release label. So it still makes sense to have a release pipeline in situations when you want to make a specially tested build. But this project doesn't have that. So I think the basic CICD should be sufficient.

build.gradle
```diff
@@ -1,9 +1,9 @@
 plugins {
     id 'java'
+    id 'com.adarshr.test-logger' version '3.2.0'
 }
 
 group 'com.surrealdb'
-version '0.1.0'
```
phughk, line 6: We definitely want to keep the version 😄 I'm not actually sure what it defaults to 🤔

build.gradle
```diff
@@ -1,9 +1,9 @@
 plugins {
     id 'java'
+    id 'com.adarshr.test-logger' version '3.2.0'
```
phughk, line 3: We don't need this plugin, do we?

build.gradle
```diff
@@ -18,9 +18,9 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.9.1'
 
     // logging
-    implementation 'org.slf4j:slf4j-api:2.0.1'
```
phughk, line 21: I saw your PR for version changes. I think we can revert the version changes here and keep the dependency issues in that other PR.

build.gradle
```diff
@@ -35,7 +35,7 @@ dependencies {
 
 test {
     useJUnitPlatform()
-    environment('TEST_SURREAL_HOST', '172.18.0.2')
+    environment('TEST_SURREAL_HOST', 'localhost')
```
phughk, line 38: Interesting. I believe this env variable is pointing to the docker container. Localhost could work if we are binding/exposing the ports. But I don't know where the previous value came from haha. Can we keep it as is and see if that works?

phughk (changes requested):
I have left some comments. There are a few things to figure out with this, but a lot of this is welcome, thank you! Will need to get the team's eye on this to see what the stance is on the CICD for the java driver. Thanks! 🙌

build.gradle
coder966: I agree, removed this commit as we have another PR for it.

.github/workflows/ci.yml
coder966: Renamed to `test`

.github/workflows/ci.yml
coder966: I am not an expert in Github Actions, but wouldn't this also run the workflow on pull requests ? And bots can still abuse it ? Anyway I applied your suggestion. This would also match other repos as you mentioned.

.github/workflows/ci.yml
coder966: Please notice the network flag, I used `host` network. But I changed it anyway to the default bridge network and mapped the port from the bridge to the host. This is to match other repos.

.github/workflows/ci.yml
coder966: I agree, because tests passing is not enough to guarantee that the build will be successful. Hence, I added the build step.

build.gradle
coder966: This make the test output more readable. But we don't need it, reverted.

build.gradle
coder966: The ip `172.18.0.2` was SurrealDB container ip on my machine. I changed it to `localhost` in order for the Github Action workflow to run. Also locahost will work see the comment above regardin the container port.

build.gradle
coder966: In `release.yml` we build and ceate a release in Gihub Release page with the same version of the tag that triggered this workflow. This means a new release will be publish automatically on new tags. Having the version hardcoded into the source code makes updaing it a manual work. However, I see your point, so please read my other comment on `release.yml`.

.github/workflows/release.yml
coder966: Thank **_you_** guys for making such an awesome database, your work makes an impact on the community and we have to give back ❤️ . 

Regarding the release mechanism, I understand this might not be the behaviour you want, but I was trying to imitate the behavior of the `surrealdb.js` repo, see: https://github.com/surrealdb/surrealdb.js/blob/main/.github/workflows/publish.yml Anyway, I will delete this workflow from this PR so that this PR can be merged.

Thanks

phughk (approved):
Thanks for the changes! Will merge in a bit.

PR #5, opened 2022-10-22
Fix SLF4J

coder966:
downgrade SLF4J to v1 because v2 is too new it is not yet supported by most frameworks.
Currently with v2 (release Aug 2022), this driver won't work on Spring framework.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,9 +18,9 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.9.1'
 
     // logging
-    implementation 'org.slf4j:slf4j-api:2.0.1'
-    testImplementation 'ch.qos.logback:logback-core:1.3.1'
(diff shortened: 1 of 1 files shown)


phughk:
Hey @coder966 ! Thanks for opening the PR and explaining the problem. I don't think we should be downgrading libraries. It may be the issue that this isn't actually SLF4J being on 2, but logback actually. Intersting that the logback downgrade works - have you tried upgrading logback in your Spring project?

coder966:
@phughk The problem was a conflict of versions of `slf4j`, since Spring brings `slf4j` v1 to the runtime classpath, and this library brings `slf4j` v2 to the runtime (and compile time) classpath. The fix is to use `compileOnly` depdendecy scope. Now the version is kept at v2.

phughk (approved):
Thanks for this @coder966 !

PR #7, opened 2022-10-28
Added namespace and database sign in

CheeseTastisch:
Added methods to signin with a namespace or database account

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,5 @@
 ## [0.1.0] - 2022-09-23
 First GA release
 
+## [0.2.0] - 2022-10-28
+Added namespace and database sign in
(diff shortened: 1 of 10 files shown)


phughk (changes requested):
Please resolve the conflicts and address the issues mentioned - will move this to draft stage until that is done. Thank you for contributing @CheeseTastisch !

CheeseTastisch:
I think it would be better to focus on pull request #6. 
As the changes in #6 seem like better improvements to me and if #6 ​​is merged this pull request will be deprecated anyway.

phughk:
It might be better, but it has so many issues due to it's size. We cannot merge it as-is even if all the issues are addressed. We need separate changes per modification. This helps track changes across versions, identify relevant changes, revert if necessary, cherry pick across versions... That PR is too big. But I agree it has a lot of merit. I will leave this up to you - if you would like to improve this solution, then it will be much faster as that other change really needs to be split properly.

PR #20, opened 2023-03-21
Replace DTOs with java records

eharrow:
Please consider this simple refactoring PR:

1. Replaces the use of the java DTO with logbook pattern with the new immutable Java Records instances.
2. Upgrades Gson to a newer version supporting Java Records when marshalling the response.
3. A fair bit of simple boilerplate code removed.
4. It possibly removes the need for the Lombok dependency.
5. Current unit tests all green.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ dependencies {
     implementation 'org.java-websocket:Java-WebSocket:1.5.3'
 
     // Gson
-    implementation 'com.google.code.gson:gson:2.9.1'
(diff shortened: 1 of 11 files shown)


phughk (commented):
Thank you for these changes @eharrow ! Personally, I am a bit averse to records. They limit functionality for a bit of convenience. Given that our implementation is in-flux, we may want to keep the capabilities of getters/setters on objects without confining ourselves to records. What do you think?

PR #26, opened 2023-03-25
Move integration tests to separate file set

eharrow:
This PR attempts to split unit tests from integration tests.  The current unit tests require a running SurrealDB instance either locally or more recently as test containers in docker which makes them integration tests by definition.  Not having used gradle before I simply adopted the standard grade pattern for integration tests with a new fileset rather than renaming them as *IT.java or something like that and use pattern matching to exclude them from unit tests.

Running `gradle check` will execute them along with checkstyle.


eharrow:
I think I might start the PR from scratch and the merge from main is too busy

eharrow:
Nightmare resolving the merge.  Closing and reopening as a new one.

Issue #27, opened 2023-03-26
Error Failed to Load SLF4J


ZeyoYT:
Hi, i am working on a project with surrealdb java lib, and i am constantly getting this error

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
```

i know about this error but i use LogBackClassic and still getting this error , here is my implementation in gradle

```
dependencies {
    ...
    implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.4.6'
    ...
    ...
    ...
   
    implementation files('build/libs/surrealdb-v0.1.0.jar')
}
```

any way to fix this?

eharrow:
I was able to reproduce easily enough.  The issue is that an older version of slf4j is being bundled in the built jar artefact.

phughk:
Hey @ZeyoYT, @eharrow fixed this and we have merged the change - closing this issue as resolved; Please re-open if it is not resolved.

PR #28, opened 2023-03-27
Fix gradle dep to exclude old slf4j and force slf4j-api:2.0.1 to be used for jar

eharrow:
This PR forces slf4j-api:2.0.1 to be used for jar else it pulls in a 1.7.X version from the websocket dependency.

https://github.com/surrealdb/surrealdb.java/issues/27 shows that an old version of slf4j is being bundled by the surrealdb driver see https://www.slf4j.org/codes.html#multiple_bindings

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ java {
 }
 
 ext {
+    slf4jVersion = "2.0.1"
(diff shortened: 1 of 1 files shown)


phughk:
Thanks @eharrow ! Good spot.

PR #29, opened 2023-03-28
Add Gradle spotless plugin

eharrow:
This PR adds the Spotless plugin to the Gradle build.  Spotless will help to keep the source formatting to a set standard and is pretty common across many projects.  I opted for the Google style as it seems pretty standard.

Also added a copy/amended version of the CONTRIBUTING.md.

This PR does not include any of the resulting source code changes from running `gradle spotlessJavaApply`.  N.B. the plugin is called as part of `gradle check` and will block the build if the source deviates from the standard.

Any way see what you think.


phughk:
Hey @eharrow, we have [checkstyle](https://github.com/surrealdb/surrealdb.java/blob/main/config/checkstyle/checkstyle.xml) to verify format. We don't need spotless even though it is a convenience. It seems like additional maintenance for something that IDEs already do.

eharrow:
Understood re check style but I cannot find a way to have it apply the formatting and instead have to rely on an IDE.

phughk:
I changed the checkstyle to google. Let's rock n roll.

PR #31, opened 2023-03-28
Move integration tests to separate file set

eharrow:
This PR attempts to split unit tests from integration tests. The current unit tests require a running SurrealDB instance either locally or more recently as test containers in docker which makes them integration tests by definition. Not having used gradle before I simply adopted the standard grade pattern for integration tests with a new fileset rather than renaming them as *IT.java or something like that and use pattern matching to exclude them from unit tests.

Running gradle check will execute them along with checkstyle.


build.gradle
```diff
@@ -66,20 +88,36 @@ spotless {
...
+    }
 }
 
 checkstyle {
     toolVersion = "10.9.3"
     configFile = resources.text.fromUri("https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${toolVersion}/src/main/resources/google_checks.xml").asFile()
 }
 
-task fatJar(type: Jar) {
+tasks.register('fatJar', Jar) {
```
eharrow, line 120: This was recommended by IntelliJ

src/test/java/com/surrealdb/connection/SurrealWebSocketConnectionTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Slf4j
+class SurrealWebSocketConnectionTest {
```
eharrow, line 14: A unit test sample with no integration.

eharrow:
@phughk any further views on this PR?  Would be nice to code up some actual unit tests as I can see a possible NPE.

phughk (approved):
This is awesome, thank you so much for this! ❤️

PR #45, opened 2023-05-16
Started to implement the Driver, Connection, Statement and ResultSet class

flipflopsen:
What did I do?
Firstly I implemented the connect method of the SurrealJDBDriver which returns a SurrealJDBCConnection with params passed through a JDBC-Connection String which gets parsed into an URI for easier usage.
Furthermore the constrcutor of SurrealJDBCConnection is implemented which connects through the Websocket and Sync-/AsnycSurrealDriver with the database to establish an authenticated connection and uses a namespace and a dbName.
In addition to this I started to implement the executeQuery()-Method of the SurrealJDBCStatement with support of arbitrary queries (with params as well (see problems section)) and the executeUpdate()-Method.
In terms of classes, the SurrealJDBCResultSet got the next().Method and the getObject(int columnIndex, Class<?> type)-Method implemented, to retrieve objects which are getting instantiated via. the gson library.

Tests
SurrealJDBCStatementTest: executeQuery, executeQueryWithArgs, executeUpdate
SurrealJDBCDriverTest: connect

Problems
The biggest problem at the moment is the retrieval of objects with arguments because the executeQuery override of the Statement only accepts a raw SQL String as an input. As a workaround I used this scheme (you can see it in the code as well):
 "select * from person where name.first = $firstName withargs [firstName, "Name"];

Another suggestion is the structure of the jdbc url: jdbc:surrealdb://host:port/dbname;namespace;optionalParam2=SomeValue2;

Would love to discuss these topics with you and appreciate any form of feedback! 

Greetings

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,6 +116,30 @@ tasks.register('integrationTest', Test) {
     }
 }
 
+tasks.register('jdbcTest', Test) {
+    description = 'Runs JDBC tests.'
+    group = 'verification'
+
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
+    shouldRunAfter test
+
+    useJUnitPlatform()
+
+    environment('TEST_SURREAL_HOST', 'localhost')
+    environment('TEST_SURREAL_PORT', 8000)
+    environment('TEST_SURREAL_USERNAME', 'root')
+    environment('TEST_SURREAL_PASSWORD', 'root')
+    environment('TEST_SURREAL_NAMESPACE', 'test')
+    environment('TEST_SURREAL_DATABASE', 'test')
+    environment('TEST_SURREAL_TOKEN', '')
+    environment('TEST_SURREAL_SCOPE', 'allusers')
+
+    testLogging {
+        events "passed"
+    }
+}
+
 checkstyle {
     toolVersion = "10.9.3"
     configFile = resources.text.fromUri("https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${toolVersion}/src/main/resources/google_checks.xml").asFile()
--- a/src/main/java/com/surrealdb/jdbc/SurrealJDBCConnection.java
+++ b/src/main/java/com/surrealdb/jdbc/SurrealJDBCConnection.java
@@ -1,5 +1,9 @@
 package com.surrealdb.jdbc;
 
+import com.surrealdb.connection.SurrealWebSocketConnection;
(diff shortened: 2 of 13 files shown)


flipflopsen:
It seems like the TestUtils class doesn't get the enviroment variables, I added a new task in the build.gradle called "jdbcTest", which is working.

Don't really now how to fix it in the pipeline, does it take the config from the build.gradle? If yes we could add the vars here:
test {
    // -- Env Vars --
    useJUnitPlatform()
}

phenzler:
> Problems
> The biggest problem at the moment is the retrieval of objects with arguments because the executeQuery override of the Statement only accepts a raw SQL String as an input. As a workaround I used this scheme (you can see it in the code as well):
> "select * from person where name.first = $firstName withargs [firstName, "Name"];

JDBC Statement does only execute an SQL query without parameters.

When using parameters a PreparedStatement must be created and the syntax of the query is defined:
Either
Select * from person where name.first = ?      (and setting values in the PreparedStatement by index)
or
Select * from person where name.first = :firstName  (and setting values in the PreparedStatement by parameter names)

flipflopsen:
Generally speaking u are right, I will move this to the PreparedStatement.
The question is now if we want to do this with '?' like normal SQL, because in the SurrealQL it's with $sth like:
"select * from person where name.first = $firstname" and then we pass a Map<String, String> as params to the Sync/Async-Driver.

The JDBC Standard isn't covering this exactly, it's more like a 
"select * from person where name.first = ?"
and then a statement.setString(1, "firstname");

phenzler:
> After reading some JDBC docs I think I've got it now

Did you finde, that it is possible to set parameters by parameter name?
Here an example:

String sql = "SELECT * FROM your_table WHERE column1 = :param1 AND column2 = :param2";
PreparedStatement pstmt = connection.prepareStatement(sql);
pstmt.setObject("param1", value1);
pstmt.setObject("param2", value2);

phughk (approved):
I gave this a quick read over and it looks really solid. Thanks for writing this Filip!

phughk:
Regarding the above discussion - we can fix things and add functionality iteratively

flipflopsen:
@phughk 
Thank you really much for reviewing!
I'll fix the test and the part with the statement as pointed out in the discussion asap so we can merge this. :)

emmanuel-keller:
Thanks for the contribution.  Since v0.2.0, the driver is now a native driver and embeds the SurrealDB SDK, redefining the way the driver is built.

Issue #50, opened 2023-06-15
REQ: Update GitHub workflow to test using integrationTest in addition to unit test


eharrow:
Merged code from a PR should pass quality checks before merging.  At present the level of testing we have is integration and not unit (we need both) however the build does not run the integration tests but merely the unit tests which are lacking and coverage.

Update test target to `gradle integrationTest` which will run the unit `test` target  before the integration tests.

eharrow:
See https://github.com/surrealdb/surrealdb.java/pull/51

phughk:
Thanks for resolving @eharrow ! 🎉

PR #51, opened 2023-06-15
chore(#50): change the build action target to integrationTest 

eharrow:
from test as unit tests do not provide sufficient coverage.  Note that this will fail quality checks as the build has a failed test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Run Tests
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: test
+          arguments: integrationTest
 
       - name: Build
         uses: gradle/gradle-build-action@v2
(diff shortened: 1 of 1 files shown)


eharrow:
Build fails as expected.  Either @phughk merge anyway or we fix the root cause - the failed test.  At least now we can have some confidence.

phughk:
This is fantastic, thank you!
Ideally I would like `gradle test` to run unit and `gradle integration` to run integration tests. The github workflow needs to change I think, but also the gradle tasks. Either of us could do that, but I am quite occupied with live queries atm. 
Will think about landing this if it isn't addressed, but if it can be addressed then that would be 10/10 amazing.
Thanks for this - we definitely need this change :)

phughk:
Great thank you! Will land this once #53 is resolved 🎉

.github/workflows/test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Run Tests
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: test
+          arguments: integrationTest
```
phughk, line 30:
```suggestion
          arguments: test integrationTest
```

phughk:
Thanks for this @eharrow ! 🎉

PR #60, opened 2023-08-14
New API proposal

phughk:
This is a proposal for a new API for the Java Driver.

The motivation behind this change (gated behind the `com.surrealdb.refactor` package) is:
- Simplify connecting. Currently, it is using a URI only, but it may use configuration in addition to the URI in the future.
- Enforced authentication before using the drivers
- Better error handling with causes and additional information
- Improved separation of concern regarding OOP
- Native SurrealDB types that can be cast
- Separation of protocol and mechanism
- Possibility to choose async or sync (via the Factory, when it is implemented)
- Possibility to add new protocols via injection (such as for embedded)

This PR is not only for review, but also for discussion about the approach. Rust has been a significant influence over this API, more so than most Java APIs, and so enums, builders, interfaces, and generics are favoured.

### DIFF
--- a/src/intTest/java/com/surrealdb/refactor/DemoScenarioTest.java
+++ b/src/intTest/java/com/surrealdb/refactor/DemoScenarioTest.java
@@ -1,0 +1,63 @@
+package com.surrealdb.refactor;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+import com.surrealdb.BaseIntegrationTest;
+import com.surrealdb.refactor.driver.BidirectionalSurrealDB;
+import com.surrealdb.refactor.driver.StatelessSurrealDB;
+import com.surrealdb.refactor.driver.SurrealDBFactory;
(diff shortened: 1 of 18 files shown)


src/intTest/java/com/surrealdb/refactor/DemoScenarioTest.java
```diff
@@ -0,0 +1,63 @@
...
+    public void testDemoScenario() throws Exception {
+        // Setup
+        URI address = getHttp().orElseThrow(() -> new IllegalStateException("No HTTP server configured"));
+        UnauthenticatedSurrealDB<BidirectionalSurrealDB> unauthenticated = new SurrealDBFactory().connectBidirectional(address);
+
+        // Authenticate
+        BidirectionalSurrealDB surrealDB = unauthenticated.authenticate(new Credentials("admin", "admin"));
+
+        // Create a multi-statement query
+        StringBuilder query = new StringBuilder("INSERT person:lamport VALUES {'name': 'leslie'};\n");
```
phughk, line 32: The new line number isn't necessary FYI. Semicolons are. We can always have our own statement builder, which could be type enforced and result in correct strings.

Issue #75, opened 2023-08-23
Feature: Handle query returning single responses in the websocket protocol


phughk:
### Is your feature request related to a problem?

The current implementation expects query to return an array of results.
If there is only 1 result, the results field will be a json object not a json array.

### Describe the solution

In the query handling in WsPlaintextConnection handle the possibility that a json object can be returned instead of an array.

### Alternative methods

None

### SurrealDB version

1.0.0-beta.9+20230821.0665e1b7 for macos on aarch64

### Contact Details

hugh@surrealdb.com

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Divine1:
@phughk please let me know if this issue is still open or completed

PR #77, opened 2023-08-25
This attempts to fix issue #75:

dshunter107:
Feature: Handle query returning single responses in the websocket protocol issue #75 

This commit incorporates allowing the user to pass a Json element that is not necessarily an array. The code will add the element (for example, a person) into a newly instantiated array if the Json element is not already an array.

Instead of throwing an error if the result is not an array, now an error is thrown if the result is not a JSON element.

The main altered code starts on line 85. I am new to adding to projects, and I am not sure why the code doesn't recognize which lines are added/changed. Instead it just says the entire file is new.

The build with gradle initially failed and gave me a prompt to run the command gradlew.bat :spotlessApply . Afterwards the program was able to build successfully with the command gradlew build.

Since I am new, can you offer me any feedback on how I did? Was this what you were looking for?

### DIFF
--- a/src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
+++ b/src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
@@ -80,8 +80,21 @@ public class WsPlaintextConnection {
                                 }
                                 JsonElement outerResultJson = resp.get("result");
                                 QueryResult[] processedOuterResults;
-                                if (outerResultJson.isJsonArray()) {
-                                    JsonArray outerResultArray = outerResultJson.getAsJsonArray();
+
+                                // checks to see if result is a JSON element, if not throw exception
+                                if (outerResultJson.isJsonObject()) {
+                                    JsonArray outerResultArray;
+
+                                    if (!outerResultJson.isJsonArray()) {
+
+                                        // add element to an array if it is not an array
+                                        outerResultArray = new JsonArray();
+                                        outerResultArray.add(outerResultJson);
+                                    } else {
+
+                                        outerResultArray = outerResultJson.getAsJsonArray();
+                                    }
+
                                     processedOuterResults =
                                             new QueryResult[outerResultArray.size()];
                                     for (int i = 0; i < outerResultArray.size(); i++) {
@@ -95,9 +108,12 @@ public class WsPlaintextConnection {
                                         processedOuterResults[i] = val;
                                     }
                                 } else {
+                                    // exception is now thrown if results are not a Json Element
+                                    // instead of Json Array
+
                                     throw new SurrealDBUnimplementedException(
                                             "https://github.com/surrealdb/surrealdb.java/issues/75",
-                                            "The response contained results that were not in an array");
+                                            "The response contained results that were not a Json Element");
                                 }
                                 return new QueryBlockResult(Arrays.asList(processedOuterResults));
                             }


phughk:
This is amazing thank you so much for contributing this! It looks like you may have accidentally switched the line endings? We should be using `LF` file endings (i.e. `\n`) instead of `\r\n` which often occurs on Windows. Could you please change that? Either in your git, or in your editor. It will make the changes easier to review and give consistency.
You might also be able to do this with `gradle spottlessApply`

dshunter107:
Thank you so much for pointing me in the right direction. I was able to change the line endings. Should I do anything else?

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
```diff
@@ -95,9 +108,12 @@ public QueryBlockResult query(String query, List<Param> params) {
                                         processedOuterResults[i] = val;
                                     }
                                 } else {
+                                    // exception is now thrown if results are not a Json Element
+                                    // instead of Json Array
+
                                     throw new SurrealDBUnimplementedException(
```
phughk, line 114: This part of the code can be turned into a protocol exception. It will be triggered if the response is an error, so technically it is unimplemented. But the ticket it is pointing to is being solve by you 🙌

phughk (commented):
This is awesome, thank you! Do you think it would be possible to have a test for it as well?
I was thinking we could have netty tests. But actually, if you want, you can move the message parsing into a separate class and then test the class with custom JSON input. Alternatively you could use the integration test file. There is a disabled test there that already works; It is disabled because netty is being too fast and connects to the db before it is initialised I think 😂

phughk:
Actually, thinking about this some more - it would be really great if there was a separate integration test for query single results 🙏 . You can use CREATE for this, as it returns a single object once created.

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
```diff
@@ -80,8 +80,21 @@ public QueryBlockResult query(String query, List<Param> params) {
...
+                                // checks to see if result is a JSON element, if not throw exception
+                                if (outerResultJson.isJsonObject()) {
+                                    JsonArray outerResultArray;
+
+                                    if (!outerResultJson.isJsonArray()) {
+
+                                        // add element to an array if it is not an array
+                                        outerResultArray = new JsonArray();
+                                        outerResultArray.add(outerResultJson);
+                                    } else {
```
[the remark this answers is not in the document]
dshunter107: okay, I am working on it now. My question would be when would the else portion of the code be run?

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
dshunter107: okay sounds good. it's up to you on which you would prefer.

dshunter107:
> Actually, thinking about this some more - it would be really great if there was a separate integration test for query single results 🙏 . You can use CREATE for this, as it returns a single object once created.

`okay. I will try this. I am rather new so it will be my first time creating a test with JUnit. Let me know how it goes. `

dshunter107:
Does SurrealDB only work with Docker?

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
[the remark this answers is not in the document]
phughk: It's a great question! The logical answer is on primitives, and null values. But the "engineering answer" is we don't know. If the `else` handles array types, and it isn't an array type, then the error we will get won't be correct. I would much rather we have clear errors where assumptions (in this case, the type) have been broken. It will make diagnosis and debugging easier. :)

phughk:
> I am rather new so it will be my first time creating a test with JUnit.
You're doing great! The way JUnit works, is there are annotations `@Test` on top of functions. When we run the test program (`gradle test` or `gradle integrationTest`), then classpath gets scanned for `@Test` functions. These then get run in such a way that output is captured separately and exceptions are caught and tracked. Then at the end of the test run you can see which tests failed.

The most common way to write tests is in the format "given, when, then". For example
```
@Test
public void test_add() {
   // given
   aNumber = 2;
   anotherNumber = 3;
   
   // when
   sum = aNumber + anotherNumber;
   
   // then
   Assertions.assertEquals(5, sum);
}
```

This test declares what is assumed to be known in the given section, then the code being tested is in the when section, and then we verify the code is correct in the `then` section.

> Does SurrealDB only work with Docker?
Towards the bottom of the readme you can set an external Docker so you don't have to rely on having it installed :) The tests also run faster that way!

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
phughk: I would like it to be a protocol exception, please :)

dshunter107:
I was able to get Docker with WSL. I was able to run the Integration tests. The error for the testSingleQueryResult test. Upon tracing the code, I am a little puzzled. It seems that the JsonQueryResultParser().parse(JsonElement) tests to see whether a "status" is a property of the Json element. 

Upon looking at QueryResult.java, I see that the "status" is attached to the Query Result (which in turn has the Json element). 

I also saw that at the current moment, only the testSingleQueryResult calls the JsonElementParser (the class I created in these commits). Is this correct? 

My question main question is should the "status" property be a part of the JsonElement or the encapuslating QueryResult? I can either put the status element into the code, or I could change the code so that it is passing in a queryResult instead of a JsonElement. 

Thanks. Sorry if this is too much of a newbie question.

src/main/java/com/surrealdb/refactor/driver/parsing/JsonElementParser.java
```diff
@@ -0,0 +1,61 @@
+package com.surrealdb.refactor.driver.parsing;
+
+import java.net.ProtocolException;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.surrealdb.refactor.exception.UnhandledProtocolResponse;
+import com.surrealdb.refactor.types.QueryResult;
+
+public class JsonElementParser {
```
phughk, line 10: Awesome separating this out, thank you!

Because the purpose of this class it to process results, and the JsonElement type isn't an important detail of the abstraction, I would instead call it something like "ResultParser".

src/main/java/com/surrealdb/refactor/driver/parsing/JsonElementParser.java
```diff
@@ -0,0 +1,61 @@
...
+	//this method parses the JSON element regardless of whether it is a single element
+	// or an array of elements
+	
+	public static QueryResult[] parseJsonElement(JsonElement outerResultJson) throws ProtocolException {
```
phughk, line 14: By having this method static there are a lot of restrictions being implied
- it cannot easily be configured via the class (say, for example, by choosing the JSON parsing implementation)
- it cannot be changed to an interface to allow multiple implementations

Furthermore, the method name contains implementation details. The implementation could likely change in the future, but we would still want to keep this method. Perhaps `parseResultMessage` would be better?

Also, I would like to avoid having explicit exceptions. I don't think explicit exception tests in the code are a nice developer experience.

phughk (commented):
Great submission, thank you! Quiet a few comments, but it's a really good change.

src/intTest/java/com/surrealdb/refactor/ConvertJavatoJson.java
```diff
@@ -0,0 +1,22 @@
+package com.surrealdb.refactor;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParser;
+
+
+
+public class ConvertJavatoJson {
```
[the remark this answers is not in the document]
dshunter107: sure

src/main/java/com/surrealdb/refactor/driver/parsing/JsonElementParser.java
dshunter107: sounds good

src/intTest/java/com/surrealdb/driver/SurrealDriverTest.java
```diff
@@ -85,29 +86,41 @@ public void testCreateWithId() {
...
     }
 
     @Test
     public void testCreateAlreadyExistsUsingUniqueIndex() {
-        assertThrows(UniqueIndexViolationException.class, () -> {
-            driver.query("DEFINE INDEX fullNameUniqueIndex ON TABLE person COLUMNS name.first, name.last UNIQUE", Collections.emptyMap(), Object.class);
-            driver.create("person", new Person("Artist", "Mia", "Mcgee", false));
-            driver.create("person", new Person("Artist", "Mia", "Mcgee", false));
-        });
+        assertThrows(
```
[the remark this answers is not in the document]
dshunter107: okay

src/intTest/java/com/surrealdb/driver/SurrealDriverTest.java
```diff
@@ -222,25 +236,44 @@ public void testDeleteAll() {
...
         Reminder insert = new Reminder("Pass this test", time);
 
         Reminder select = driver.create("reminder", insert);
         assertNotNull(select.getTime());
         assertEquals(time, select.getTime());
-	}
+    }
+
+    @Test
+    public void testSingleQueryResult() {
```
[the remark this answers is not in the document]
dshunter107: oh wow. Good to know. Thanks! will do.

src/intTest/java/com/surrealdb/refactor/ConvertJavatoJson.java
```diff
@@ -0,0 +1,22 @@
...
+
+	public static JsonElement convertObject(Object object) {
```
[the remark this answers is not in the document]
dshunter107: I understand about the gson to Json.

 Interesting, so the JsonParseElement (which will be renamed) will handle the status and errors. Thanks for the clarification.

src/main/java/com/surrealdb/refactor/driver/parsing/JsonElementParser.java
dshunter107: sounds good. I will implement the changes. Thanks for all the feedback. I feel like I am learning a lot.

phughk: That's awesome to hear! Really happy about that :) It's great code overall, so I really hope the feedback isn't off-putting. It's getting close to being able to land.

phughk:
Have a look at this test to see what an integration test would look like for the refactored code
https://github.com/surrealdb/surrealdb.java/blob/main/src/intTest/java/com/surrealdb/refactor/DemoScenarioTest.java
The goal is to move away from the original implementation to better support types, protocols, lifecycle, and various implementations and configuration.

src/intTest/java/com/surrealdb/driver/SurrealDriverTest.java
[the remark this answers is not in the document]
dshunter107: I am using gradlew.bat spottlessApply (As gradle is not recognized in the project folder directory), and I am even making sure to convert delimiters on eclipse. My settings are even such that only the proper delimiters should be implemented on new code, but this still happens.

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
```diff
@@ -0,0 +1,63 @@
...
+                }
+
+                QueryResult val = new JsonQueryResultParser().parse(innerResultJson);
+                processedOuterResults[i] = val;
+            }
+        } else {
+
+            throw new UnhandledProtocolResponse(
+                    "https://github.com/surrealdb/surrealdb.java/issues/75\n"
+                            + "The response contained results that were not a Json Element");
```
phughk, line 59: Only the unimplemented exception should have links. This is because the unimplemented exception is known to be unimplemented, so we know it should have a ticket. Unhandled input is a real edge case that we shouldn't see - it would mean that either the protocol is broken or that we have a bug.

```suggestion
            throw new UnhandledProtocolResponse("Expected the input JSON for handling results to be a JSON Object");
```

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
```diff
@@ -0,0 +1,63 @@
...
+            } else if (resultMessage.isJsonArray()) {
+
+                outerResultArray = resultMessage.getAsJsonArray();
+            } else {
+                // this puts in the else if for the if it is a jsonArray; however,
+                // I am unsure of when this portion of the code would run
+                throw new RuntimeException(
+                        "https://github.com/surrealdb/surrealdb.java/issues/75\n "
+                                + "The Outer Result Array came up for false for (1) is not an"
+                                + " Array and (2) is an Array");
```
phughk, line 42: We don't need URLs, and the message can be clearer :) When users (or driver developers) see the exception, they need to be able to see what happened conveniently - implementation details tend not to be important, only explaining the assumptions and why things are having to "break" with an exception.

```suggestion
                throw new RuntimeException("The input json message was neither an object or an array");
```

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
```diff
@@ -0,0 +1,63 @@
...
+    public ResultParser() {}
+
+    public QueryResult[] parseResultMessage(JsonElement resultMessage) {
+        QueryResult[] processedOuterResults;
+
+        // checks to see if result is a JSON element, if not throw exception
+        if (resultMessage.isJsonObject()) {
+            JsonArray outerResultArray;
+
+            if (!resultMessage.isJsonArray()) {
```
phughk, line 22: You have already tested that it is an object above :) So we don't need to test that it is not an array here - we already know it isn't

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
```diff
@@ -81,15 +79,13 @@ public QueryBlockResult query(String query, List<Param> params) {
                                 }
                                 JsonElement outerResultJson = resp.get("result");
                                 QueryResult[] processedOuterResults;
-								
-                                
+                                ResultParser resultParser = new ResultParser();
```
phughk, line 82: We can have the result parser as a class property. That way it can be injected for "production" with a correct implementation, but have a "fake" implementation injected for tests, such as a mock implementation. It also reduces garbage collection.

phughk (commented):
A lot of good progress here! Left some more comments

> should the singleQueryResultTest be inside DemoScenarioTest.java ?

Yes please! That way we have 2 scenarios covered. We can always move the test out when necessary.

> Also, earlier you mentioned using the driver.create(Object) function. With the DemoScenarioTest class, I am not sure if there is a similar function.

You shouldn't use the create function, actually - we only have `query` supported in the rewrite currently. So we should use the `query` method and have a `CREATE` SurrealQL statement - https://surrealdb.com/docs/surrealql/statements/create

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
dshunter107: I didn't think about that. very true!

dshunter107:
I have a question about the representative JSonObject. Person has a nested object "Name". In order to have the JsonParseRequest accept the person, I had to put only primitives in the Json file. As a result the person is represented by 
title: "developer" 
first: "David"
last: "Hunter"
'
My question is whether this is okay or not. The representative Json object could be like the following: 
title: "Developer"
name.first: "David"
name.last: "Hunter"

I think it might make a difference depending on what you are interested on doing in the future. Let me know what your thoughts are or if it even matters

phughk:
> I have a question about the representative JSonObject. Person has a nested object "Name". In order to have the JsonParseRequest accept the person, I had to put only primitives in the Json file.

The Value object doesn't yet handle all supported types. We can add them later. But at the moment we want to handle single response alongside multiple response. So it's ok if it is a flat object instead of a nested object.

src/intTest/java/com/surrealdb/driver/SurrealDriverTest.java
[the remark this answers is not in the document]
phughk: Yeah that's fine then. The project doesnt have spotless autoapplied and clearly the checksyle check doesn't work.... it's fine, ignore it. Please try to remember to do the spotless apply before pushing and we should be good I think 🙏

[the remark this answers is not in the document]
dshunter107: sounds good

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
dshunter107: okay. done

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
dshunter107: done

src/intTest/java/com/surrealdb/driver/SurrealDriverTest.java
```diff
@@ -222,25 +236,44 @@ public void testDeleteAll() {
...
+        int expectedSize = 1;
+
+        Person singlePerson =
```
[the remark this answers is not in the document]
dshunter107: okay.

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
dshunter107: I put it as a class property

dshunter107:
> > I have a question about the representative JSonObject. Person has a nested object "Name". In order to have the JsonParseRequest accept the person, I had to put only primitives in the Json file.
> 
> The Value object doesn't yet handle all supported types. We can add them later. But at the moment we want to handle single response alongside multiple response. So it's ok if it is a flat object instead of a nested object.

Okay. Then I think everything should be okay. Could you let me know if there is anything you would like me to change?

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
```diff
@@ -0,0 +1,84 @@
...
+            preparedJsonObject.addProperty("status", "ok");
+            preparedJsonObject.addProperty("time", System.currentTimeMillis());
+            preparedJsonObject.add("result", innerResultArray);
+
+            // add prepared object to an array
+            outerResultArray = new JsonArray();
+            outerResultArray.add(preparedJsonObject);
+
+        } else if (resultMessage.isJsonArray()) {
+
```
phughk, line 43:
```suggestion
```

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map.Entry;
+import java.util.Set;
+
+public class ResultParser {
+    // this method parses the JSON element regardless of whether it is a single element
+    // or an array of elements
+
+    public ResultParser() {}
+
+    public QueryResult[] parseResultMessage(JsonElement resultMessage) {
```
phughk, line 17: Tbh we can have this be JsonObject instead of JsonElement. The check should be done outside the function. It means we have less edge cases and less incorrect usage.

phughk (commented):
Looks mostly fine. If you could apply the changes recommended and change the method signature that would be fab!

I will likely do some touch ups once it is merged, but never the less thank you very much for this!

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
```diff
@@ -0,0 +1,84 @@
...
+
+            // add prepared object to an array
+            outerResultArray = new JsonArray();
+            outerResultArray.add(preparedJsonObject);
+
+        } else if (resultMessage.isJsonArray()) {
+
+            outerResultArray = resultMessage.getAsJsonArray();
+        } else {
+
```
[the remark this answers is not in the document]
dshunter107: okay

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
dshunter107: ok

src/main/java/com/surrealdb/refactor/driver/parsing/ResultParser.java
dshunter107: I turned them into separate functions: 1 from JsonElement, 1 from JsonObject, and 1 from JsonArray

PR #78, opened 2023-09-14
Use modern Java and cleanups

JFortun:
This PR is a refactor of the project to:

- Fix some common issues
- Remove unused `import`s
- Use immutable `record`s instead of classes when possible and avoid lombok to make code more readable and shorter
- Added a ton of missing `final` keywords when possible
- Run `gradle :spotlessApply` to correct formatting
- Use lambdas in a lot of places when possible to reduce the complexity of the code


phughk:
Hey thanks! Will merge this once conflicts are resolved! 🎉

JFortun:
> Hey thanks! Will merge this once conflicts are resolved! 🎉

Hello @phughk, conflicts resolved.

JFortun:
@phughk Can you assign this PR to me? I don't have access.

phughk:
Actually just gave it a proper read (yesterday was at the hacktoberfest event so split attention). I thought this was just spotless apply - not sure why variables inside blocks are being made final? They are "final" in a way, but constantly writing final is terse and we don't need the keyword because it isn't exposed beyond the function. So it is unnecessary and having `final` made explicit may give the wrong impression that its deliberate (when I believe what you mean is that it isn't modified i.e. implicit).
Could you revert the final keywords please?

JFortun:
> Actually just gave it a proper read (yesterday was at the hacktoberfest event so split attention). I thought this was just spotless apply - not sure why variables inside blocks are being made final? They are "final" in a way, but constantly writing final is terse and we don't need the keyword because it isn't exposed beyond the function. So it is unnecessary and having `final` made explicit may give the wrong impression that its deliberate (when I believe what you mean is that it isn't modified i.e. implicit).
> 
> Could you revert the final keywords please?
> 
> RE assigning - assigning is for who owns the review. Will assign to me.

Thanks for the review @phughk.

The final keywords serve a purpose, even if the variable is not exposed beyond the function, the JVM is able to do some optimizations when we set a varible as final.
[Here you have some benchmarks and info about the advantages](https://www.baeldung.com/java-final-performance#1-performance-test)

About assigning, that's not for the one that does the review. Who owns the review goes in the "reviewers" field just above.
The PRs are assigned to the ones working on them. Each field has their purpose.

phughk:
Based on the 2 answers in SO (not docs), people say that assignees are the "stewards" or go to owners of the PR lifecycle (rejected, timely reviews, merging etc).
https://stackoverflow.com/questions/41087206/on-github-whats-the-difference-between-reviewer-and-assignee
That's how I've been using it so far. The reviewers are just a flexible list of people who need to or have reviewed.

Thanks for sharing about the final stuff I had no idea. Makes a lot of sense :) Re-reviewing now

.editorconfig
```diff
@@ -14,7 +14,6 @@ trim_trailing_whitespace = true
 
 [*.{yml,json}]
 indent_size = 2
-indent_style = space
```
phughk, line 17: Why was this removed? We do want 2 spaces

CONTRIBUTING.md
```diff
@@ -1,6 +1,6 @@
 <h1><img height="25" src="/img/contributing.svg">&nbsp;&nbsp;Contributing</h1>
 
-We would &nbsp;<img width="15" src="/img/love.svg">&nbsp; for you to contribute to SurrealDB.java and help make it better! We want contributing to SurrealDB to be fun, enjoyable, and educational for anyone and everyone. All contributions are welcome, including features, bugfixes, and documentation changes, as well as updates and tweaks, blog posts, workshops, and everything else.
+We would &nbsp;<img width="15" src="/img/love.svg">&nbsp; for you to contribute to SurrealDB.java and help make it better! We want to contribute to SurrealDB to be fun, enjoyable, and educational for anyone and everyone. All contributions are welcome, including features, bugfixes, and documentation changes, as well as updates and tweaks, blog posts, workshops, and everything else.
```
phughk, line 3: Please revert, it was correct. This is also a file managed across repositories so I would change in surrealdb/surrealdb since that is the parent file that gets copied over.

README.md
```diff
@@ -104,8 +104,8 @@ Integration tests depend on having docker installed.
 
 The following environment variables can be used to avoid relying on docker.
 
-| ENV | Description |
-|--|--|
+| ENV                 | Description                   |
+|---------------------|-------------------------------|
```
phughk, line 108: Tbh I prefer the original. It doesn't look as good, but it won't need to be maintained as much. It renders to the same look.

src/intTest/java/com/surrealdb/driver/SurrealDriverSpecialOperationsTest.java
```diff
@@ -29,93 +23,92 @@ public class SurrealDriverSpecialOperationsTest extends BaseIntegrationTest {
...
+                new SurrealWebSocketConnection(testHost, testPort, false);
         connection.connect(5);
-        driver = new SyncSurrealDriver(connection);
+        this.driver = new SyncSurrealDriver(connection);
     }
 
     @Test
     public void testSignIn() {
-        driver.signIn(TestUtils.getUsername(), TestUtils.getPassword());
+        this.driver.signIn(TestUtils.getUsername(), TestUtils.getPassword());
```
phughk, line 34: We don't need to constantly specify `this`. Please revert - I would like `this` to only be used in ambigous situations. There shouldnt be a driver variable or parameter in this code and if there is, it should have a different name.

src/main/java/com/surrealdb/jdbc/SurrealJDBCConnection.java
```diff
@@ -21,272 +19,287 @@
 
 public class SurrealJDBCConnection implements Connection {
     @Override
-    public Statement createStatement() throws SQLException {
+    public Statement createStatement() {
```
phughk, line 22: The exceptions I think are part of the interface, no? So should still be part of the signature.

src/main/java/com/surrealdb/jdbc/SurrealJDBCConnection.java
phughk, line 1: Could you please revert this entire file. It didnt need refactoring as it was automatically generated from the interface. That goes for exceptions, method order, signatures, formatting etc.

src/main/java/com/surrealdb/refactor/driver/HttpConnection.java
```diff
@@ -3,59 +3,50 @@
...
-                                        "HTTP connections are not yet implemented");
-                            }
-                        };
-                return new UnusedSurrealDB<>() {
-
-                    @Override
-                    public StatelessSurrealDB use() {
+        return credentials -> {
+            final StatelessSurrealDB surrealdb =
+                    (query, params) -> {
```
phughk, line 22: Please revert - this isn't "unused" code, its stubbed for further development

src/main/java/com/surrealdb/refactor/driver/HttpConnection.java
phughk, line 1: Could you please revert this entire file. The `final` changes can stay, but this is a work in progress file with a lot of stubbed out code. It doesn't need to optimised.

src/main/java/com/surrealdb/refactor/driver/StatelessSurrealDB.java
```diff
@@ -9,8 +9,8 @@ public interface StatelessSurrealDB {
     /**
      * Perform a general purpose query against the database.
      *
-     * @param query
-     * @return
+     * @param query the query to run
+     * @return a QueryBlockResult
```
phughk, line 13:
```suggestion
     * @param query the SurrealQL String query to run
     * @return a QueryBlockResult representing several results possible for each statement
```

src/main/java/com/surrealdb/refactor/driver/SurrealDBWebsocketClientProtocolHandler.java
```diff
@@ -24,42 +27,41 @@ public class SurrealDBWebsocketClientProtocolHandler
...
-
     private Channel channel;
 
     public SurrealDBWebsocketClientProtocolHandler() {}
 
     @Override
-    public void handlerAdded(ChannelHandlerContext ctx) {
-        handshakeFuture = ctx.newPromise();
+    public void handlerAdded(final ChannelHandlerContext ctx) {
+        this.handshakeFuture = ctx.newPromise();
```
phughk, line 38: Same comment about `this` - we don't want to keep typing it, so please revert

src/main/java/com/surrealdb/refactor/driver/WsPlaintextConnection.java
```diff
@@ -28,38 +34,41 @@ public class WsPlaintextConnection {
...
         System.out.printf("Connecting to %s\n", uri);
         try {
             channel = bootstrapProtocol(uri).connect(uri.getHost(), uri.getPort()).sync().channel();
-        } catch (InterruptedException e) {
+        } catch (final InterruptedException e) {
             throw new RuntimeException(e);
         }
 
-        return new UnauthenticatedSurrealDB<BidirectionalSurrealDB>() {
+        return new UnauthenticatedSurrealDB<>() {
```
phughk, line 49: Please revert. In the future we are going to turn these into interfaces or abstract classes and it will be good to have the signatures explicit.

src/main/java/com/surrealdb/refactor/package-info.java
```diff
@@ -1,21 +1,14 @@
...
- * - correct types
- * - simple to navigate, test, and mock
- *
- * This change won't be a drop in replacement for the original driver, so treat this as experimental until
- * it does replace the original implementation.
- *
- * This interface is extremely likely to change. Primarily, I don't want to expose internal implementation details.
- * As I am writing this, I am aware this is happening for a lot of the code.
- * It will need refactoring to prevent exposing details.
+ * @author surrealdb The refactor package contains the new implementation of the driver. The goal is
```
phughk, line 2: Why did this change? Please revert

src/main/java/com/surrealdb/refactor/types/Credentials.java
phughk, line 1: Please revert this entire file

phughk (commented):
Heya so I reviewed a bit more properly now. Thank you for writing it. I do have some reservations. I think it would be good to split this up.

I think we can have the following PRs
- introduce final
- make imports explicit and spotless apply
- rearrange declaration of interfaces etc

That will make it easier to review and make sure nothing unwanted slips in.

What i would like to avoid
- explicit `this` - its unnecessary and we don't want to constantly write it
- markdown, javadoc modifications - these are largely unnecessary. If we are adding them, I would like them to be meaningful.
- removing parameters/properties interfaces etc - we want to keep them, especially for the refactor (rewrite) package, since that is work in progress and people need to understand the direction things are going
- jdbc stuff - please dont touch it. It was auto-generated and doesnt need to be fixed on the surface level.

But it was good to see where things can be improved and like I say - this can be several changes that would be easy to accept. Thank you!

.editorconfig
JFortun: It's removed because the `indent_style` property with the same value is already defined in the `[*]` section so it's redundant and not applied.

README.md
JFortun: Reverted.

CONTRIBUTING.md
JFortun: Reverted.

src/intTest/java/com/surrealdb/driver/SurrealDriverSpecialOperationsTest.java
JFortun: Removed all the `.this`.

src/main/java/com/surrealdb/refactor/driver/SurrealDBWebsocketClientProtocolHandler.java
JFortun: Removed all the `.this`.

src/main/java/com/surrealdb/refactor/types/Credentials.java
JFortun: Even if this file is part of an in-progress refactor, it does not give any advantage to have it as a full class when the record it's the same (and with immutability advantages).
>we want to keep them, especially for the refactor (rewrite) package, since that is work in progress and people need to understand the direction things are going

If the direction is to have boilerplate when it's not needed maybe we have to rethink the direction (just a proposal, your repo = your rules).

Issue #82, opened 2024-03-18
Feature: A complete SDK With Repository pattern


Mukund2900:
### Is your feature request related to a problem?

I want to use this sdk with springboot and need feature like findById and other features which can help me retrieve data directly into the specified DTO. 
In the current scenario the only way to do this is to run a select query and expect list of items and then check if size is 1 or not in the case where i want to fetch by id. 

### Describe the solution

A similar interface like data jpa where we can run queries and autowire that class. 

### Alternative methods

-

### SurrealDB version

1.3.0 for macos on aarch64

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

naisofly:
Hey @Mukund2900! 

While we can do that, they'd be client-side sugar on top of what you could do yourself. 🍭

No worries though, we're always open to new feature requests! 🙌 Just keep in mind that repo patterns are primarily a client concern, so it's not a top priority for now with our regular [releases](https://surrealdb.com/releases). But we'll definitely keep it on the radar!

In the meantime, feel free to share any other ideas you've got brewing. We're all ears 👂🏽😁

Mukund2900:
@naisofly , I kindly request your review of this pull request. It represents just the beginning, with ample opportunity for further enhancements. I welcome your insights and suggestions on the same.

PR #83, opened 2024-03-26
Initial commit for SDK With Repository pattern

Mukund2900:
#82 
I raised this feature request a couple of days back. 
As you guys mentioned that you are open to ideas, here is an approach which can help spring developers to easily integrate surrealdb with their existing system. 

I have created an implementation which will expect the following to use repository pattern  (The library still works as it is if this is not required) ->

1.  Configuration through application.properties (else it will take the default values, will add id and password as well) :

> surrealdb.host=localhost
> surrealdb.port=8000
> surrealdb.useSsl=false
> surrealdb.nameSpace=test
> surrealdb.database=test

2.  Entity Annotation with @SurrealTable and @SurrealId: Entities can now be annotated to define document/table mappings and identifiers, simplifying the mapping process. For example:

```
@SurrealTable("parent_route")
@AllArgsConstructor
@Data
@NoArgsConstructor
public class ParentRoute {

    @SurrealId
    private String id;
    private Long validFrom;
    private Long validTo;
    private String client;
    private String rId;

 } 
```

3. CRUD Operations through SurrealCrudRepository: Entities can extend the SurrealCrudRepository interface to enable CRUD operations, streamlining data manipulation tasks:

```
@Repository
public interface ParentRouteRepository extends SurrealCrudRepository<ParentRoute,String> {
}
```

4. Custom Queries with @SurrealQuery: Similar to JPA's @Query annotation, @SurrealQuery enables the execution of custom queries, providing flexibility in data retrieval:

```
@Repository
public interface ParentRouteRepository extends SurrealCrudRepository<ParentRoute,String> {

    @SurrealQuery("select * from parent_route where client = ?1 and rId = ?2 ;")
    ParentRoute findByRouteIdAndBuid(String client, String rId);
    
    
    @SurrealQuery("select *, ->hasData->(r_data where time = ?1 ).*->hasVehicle->vehicle_detail.* as vehicle from parent_route where client = ?2 and rId = ?3 ;")
    List<ParentRoutePrj> findShuttleRouteByStartTimeAndBuidAndRouteId(long time, String client, String rId);
    
}
```

Extra feature in first cut -> 
If we want a field to have a sequence we can use @SurrealSequenceId(sequenceName = "rIdSequence") : This will give that field a numeric sequence (with string-type fields being appropriately casted).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ plugins {
 }
 
 group 'com.surrealdb'
-version '0.1.0'
+version '0.1.1'
 
 repositories {
     mavenCentral()
@@ -51,6 +51,10 @@ dependencies {
 
     // Gson
     implementation 'com.google.code.gson:gson:2.9.1'
+
+    //spring
+    implementation 'org.springframework:spring-webmvc:5.3.20'
+    implementation 'org.reflections:reflections:0.10.2'
 
     // logging
     implementation "org.slf4j:slf4j-api:${slf4jVersion}"
--- a/src/main/java/com/surrealdb/config/RepositoryInitializer.java
+++ b/src/main/java/com/surrealdb/config/RepositoryInitializer.java
@@ -1,0 +1,116 @@
+package com.surrealdb.config;
+
+import com.surrealdb.driver.model.QueryResult;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.reflections.Reflections;
+import org.reflections.scanners.SubTypesScanner;
+import org.reflections.util.ClasspathHelper;
+import org.reflections.util.ConfigurationBuilder;
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.beans.factory.support.*;
+import org.springframework.context.annotation.Configuration;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.sql.SQLException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.surrealdb.config.SurrealDBConnection.getRepoDriver;
+
+@Configuration
+public class RepositoryInitializer implements BeanDefinitionRegistryPostProcessor {
+    public Set<Class<? extends SurrealCrudRepository>> findSurrealCrudRepositories(String packageName) {
+        Reflections reflections = new Reflections(new ConfigurationBuilder()
+            .setUrls(ClasspathHelper.forPackage(packageName))
+            .setScanners(new SubTypesScanner(false)));
+
+        return reflections.getSubTypesOf(SurrealCrudRepository.class);
+    }
+
+    @Override
+    public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) throws BeansException {
+        Set<Class<? extends SurrealCrudRepository>> allClasses = findSurrealCrudRepositories("");
+        for (Class<? extends SurrealCrudRepository> repoClass : allClasses) {
+            Type[] genericInterfaces = repoClass.getGenericInterfaces();
+            for (Type genericInterface : genericInterfaces) {
+                if (genericInterface instanceof ParameterizedType) {
+                    ParameterizedType parameterizedType = (ParameterizedType) genericInterface;
+                    if (parameterizedType.getRawType().equals(SurrealCrudRepository.class)) {
+                        Type[] typeArguments = parameterizedType.getActualTypeArguments();
+                        if (typeArguments.length == 2) {
+                            Type entityType = typeArguments[0];
+                            Class<?> entityClass = null;
+                            if (entityType instanceof Class<?>) {
+                                entityClass = (Class<?>) entityType;
+                            }
+                            registerRepositoryBean(entityClass, registry, repoClass);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private <T, ID, R extends SurrealCrudRepository<T, ID>> void registerRepositoryBean(Class<T> entityType, BeanDefinitionRegistry registry, Class<R> repositoryClass) {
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
 }
 
 group 'com.surrealdb'
-version '0.1.0'
+version '0.1.1'
```
phughk, line 11: This wouldnt be a patch, and we would version into 0.2.0 independently when we do a release
```suggestion
version '0.1.0'
```

build.gradle
```diff
@@ -52,6 +52,10 @@ dependencies {
     // Gson
     implementation 'com.google.code.gson:gson:2.9.1'
 
+    //spring
+    implementation 'org.springframework:spring-webmvc:5.3.20'
```
phughk, line 56: By having these dependencies in the project, it means that the driver is tied with the spring. We don't want people to be forced to use spring.

It makes sense for such repository code to be in a separate module, that the spring users can include, but the non-spring users don't need to include.

src/main/java/com/surrealdb/config/RepositoryInitializer.java
```diff
@@ -0,0 +1,116 @@
...
+        ProxyFactory proxyFactory = new ProxyFactory();
+        proxyFactory.addInterface(repositoryClass);
+        SurrealCrudRepositoryImpl<T, ID> target = new SurrealCrudRepositoryImpl<>(entityType);
+
+        proxyFactory.addAdvice((MethodInterceptor) invocation -> {
+            Method method = invocation.getMethod();
+            if (method.isAnnotationPresent(SurrealQuery.class)) {
+                String query = method.getAnnotation(SurrealQuery.class).value();
+                Object[] args = invocation.getArguments();
+                Pattern pattern = Pattern.compile("\\?(\\d+)");
```
phughk, line 71: This could be static outside the method

src/main/java/com/surrealdb/config/RepositoryInitializer.java
```diff
@@ -0,0 +1,116 @@
...
+                Matcher matcher = pattern.matcher(query);
+                StringBuffer sb = new StringBuffer();
+                while (matcher.find()) {
+                    int index = Integer.parseInt(matcher.group(1)) - 1;
+                    if (index < args.length) {
+                        String replacement = args[index] instanceof String ? "'" + args[index] + "'" : args[index].toString();
```
phughk, line 77: You can use String.format here

src/main/java/com/surrealdb/config/SurrealCrudRepository.java
```diff
@@ -0,0 +1,12 @@
+package com.surrealdb.config;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface SurrealCrudRepository<T, ID> {
```
phughk, line 6: This is a nice convenience, but it goes beyond what the driver is.

That said the driver in Rust does provide a DSL-like experience, so this fits under that.

In order not to pollute the driver codebase further, I would recommend having such patterns outside the driver in a separate repository. Otherwise this becomes bundled with the driver and we have the burden of maintaining it - providing variations and fitting it into all manner of APIs. That isn't the purpose of the driver.

src/main/java/com/surrealdb/config/SurrealCrudRepositoryImpl.java
```diff
@@ -0,0 +1,142 @@
...
+    private static String resolveTableName(Class<?> entityType) {
+        if (!entityType.isAnnotationPresent(SurrealTable.class)) {
+            throw new MissingSurrealTableAnnotationException(entityType);
+        }
+        return entityType.getAnnotation(SurrealTable.class).value();
+    }
+
+    @Override
+    public Optional<T> findById(ID id) {
+        String query = String.format("select * from %s where id = '%s';", tableName, id.toString());
```
phughk, line 35: This is a security vulnerability - see "sql injection". It happens when you take input and put it into a trusted query. You should instead use the protocol-level parameter handling.

src/main/java/com/surrealdb/config/SurrealCrudRepositoryImpl.java
```diff
@@ -0,0 +1,142 @@
...
+                        }
+                    }
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Failed to access or set value for @SurrealSequenceId field", e);
+                }
+            }
+        }
+    }
+
+    private Long generateRouteId(String sequenceName) {
```
phughk, line 93: I don't know what this is but it looks wrong? SurrealDB doesnt provide sequences and creating them artificially is misleading, incorrect, and prone to unexpected bugs.

src/main/java/com/surrealdb/config/SurrealCrudRepositoryImpl.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public T update(T entity) {
+        return driver.update(getIdFromEntity(entity), entity).get(0);
+    }
+
+    private String getIdFromEntity(T entity) {
+        for (Field field : entity.getClass().getDeclaredFields()) {
+            if (field.isAnnotationPresent(SurrealId.class)) {
```
phughk, line 125: Overall, I'm not opposed to having annotation handling like this. Would be willing to accept annotation handling without spring

phughk (commented):
Thank you so much for opening this PR!

This goes beyond the scope of what the driver should do. We could definitely see such helpers working for a separate jar/module though! There are certainly a lot of spring users out there. However, we need to cater to all jvm users, not just spring users. So we cannot have
- spring as a dependency
- patterns that tie us to a particular technology (in this repo)
- add scope creep to the driver that users would then rely on

It would definitely be useful to have such code in a different repository though.

phughk:
Closing this PR as we have agreed with @Mukund2900 it is better in a separate repository

Mukund2900:
I have created a seperate repo for [this](https://github.com/Mukund2900/surrealdb-spring-client).

naisofly:
Thanks @Mukund2900 ! Feel free to contribute to the Awesome Surreal repo, perhaps under [Client Libraries](https://github.com/surrealdb/awesome-surreal/?tab=readme-ov-file#client-libraries) ?

Issue #90, opened 2024-10-12
Bug: java.lang.UnsatisfiedLinkError


047massi:
### Describe the bug

I encountered a java.lang.UnsatisfiedLinkError when calling the queryBind method.

### Steps to reproduce

`surreal.queryBind("SELECT * FROM table WHERE id = $id", paramsMap);`

### Expected behaviour

The `queryBind` method should successfully execute a parameterized query on the database without throwing any errors. It should bind the specified parameters to the query and return the expected result set or a successful response

### SurrealDB version

surreal 2.0.4

### Contact Details

m46955611@gmail.com

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

odwrotnie:
Same problem here :(

i1Fury:
Same problem. Any plans on continuing work on this repository?

`response = this.driver.queryBind(
                    "UPSERT $what SET $object;",
                    Map.of("what", what, "object", object)
            );`

`Caused by: java.lang.UnsatisfiedLinkError: 'long com.surrealdb.Surreal.queryBind(long, java.lang.String, java.util.Map)'`

masterofdaemon:
something similar :   `No implementation found for boolean com.surrealdb.Entry.deleteInstance(long) (tried Java_com_surrealdb_Entry_deleteInstance and Java_com_surrealdb_Entry_deleteInstance__J) - is the library loaded, e.g. System.loadLibrary?
2025-10-04 01:02:40.854 14717-14725 System                  com.example.kmp                      E  Uncaught exception thrown by finalizer
2025-10-04 01:02:40.854 14717-14725 System                  com.example.kmp                      E  java.lang.UnsatisfiedLinkError: No implementation found for boolean com.surrealdb.Entry.deleteInstance(long) (tried Java_com_surrealdb_Entry_deleteInstance and Java_com_surrealdb_Entry_deleteInstance__J) - is the library loaded, e.g. System.loadLibrary?
                                                                                                    	at com.surrealdb.Entry.deleteInstance(Native Method)
                                                                                                    	at com.surrealdb.Native.deleteInstance(Native.java:25)
                                                                                                    	at com.surrealdb.Native.finalize(Native.java:58)
                                                                                                    	at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:339)
[...]

PR #106, opened 2025-02-12
Impl queryBind

### LINKED ISSUE #98: Bug: queryBind causes premature termination in a exists method
### Describe the bug

In the exists method, the process terminates prematurely after the execution of queryBind. All subsequent code, including logging ("exist?: ...") and the return statement, is never executed. This issue only occurs when using .queryBind to execute the database query.

### Steps to reproduce

```java
public static boolean exists(String username, String email) {
    logger.info("step: 1");
    Response response = surreal().queryBind(
        "SELECT * FROM users WHERE username = $username AND email = $email;", 
        Map.of("username", username, "email", email)
    );
    logger.info("step: 2");

    boolean exist = response.take(User.class, 0) != null;
    logger.info("exist?: " + exist);

    return exist;
}
```

### Expected behaviour

The .queryBind method should not block or terminate the process. The subsequent code should execute correctly, logging whether the user exists and returning a boolean value.

### SurrealDB version

surreal 2.1.4

### Contact Details

m46955611@gmail.com

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

harry-xi:
Fixes #98

### DIFF
--- a/src/main/rust/surreal.rs
+++ b/src/main/rust/surreal.rs
@@ -205,7 +205,7 @@ pub extern "system" fn Java_com_surrealdb_Surreal_query<'local>(
 }
 
 #[no_mangle]
-pub extern "system" fn Java_com_surrealdb_Surreal_queryWithValue<'local>(
+pub extern "system" fn Java_com_surrealdb_Surreal_queryBind<'local>(
     mut env: JNIEnv<'local>,
     _class: JClass<'local>,
     ptr: jlong,
--- a/src/main/rust/surreal.rs
+++ b/src/main/rust/surreal.rs
@@ -219,13 +219,13 @@ pub extern "system" fn Java_com_surrealdb_Surreal_queryBind<'local>(
     let query = get_rust_string!(&mut env,&query, || 0);
     let keys = get_rust_string_array!(&mut env, params_keys, || 0);
     let value_ptrs = get_long_array!(&mut env, &params_values, || 0);
-    let mut parems_map = BTreeMap::<String,&Value>::new();
+    let mut params_map = BTreeMap::<String,&Value>::new();
     for (key,value_ptr) in keys.into_iter().zip(value_ptrs) {
         let value = get_value_mut_instance!(&mut env, value_ptr, || 0);
-        parems_map.insert(key, value);
+        params_map.insert(key, value);
     }
     
-    let res = surrealdb_query::<&Value>(&surreal, &query, Some(parems_map));
+    let res = surrealdb_query::<&Value>(&surreal, &query, Some(params_map));
     let res = check_query_result!(&mut env, res, || 0);
     // Build a response instance
     JniTypes::new_response(Arc::new(Mutex::new(res)))
--- a/src/main/rust/surreal.rs
+++ b/src/main/rust/surreal.rs
@@ -207,6 +207,57 @@ pub extern "system" fn Java_com_surrealdb_Surreal_query<'local>(
     JniTypes::new_response(Arc::new(Mutex::new(res)))
 }
 
+#[no_mangle]
+pub extern "system" fn Java_com_surrealdb_Surreal_queryWithValue<'local>(
+    mut env: JNIEnv<'local>,
+    _class: JClass<'local>,
+    ptr: jlong,
+    query: JString<'local>,
+    params_keys: JObjectArray<'local>,
+    params_values: JLongArray<'local>,
+) -> jlong {
+    // Retrieve the Surreal instance
+    let surreal = get_surreal_instance!(&mut env, ptr, || 0);
+    // Retrieve the query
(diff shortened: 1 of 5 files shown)


harry-xi:
I found a little question: this implement doesn't support the value to be `null`. Is this important?

massikhr:
I noticed that `Response.take(Class<T>, int)` throws a` NullPointerException` if the query returns no results instead of returning null or an empty instance. This makes error handling more difficult in cases where no matching entry exists.

Would it be possible to handle this case more gracefully by returning `null` or an empty object instead of throwing an exception?

Here is my stack trace when this issue occurs:

`java.lang.NullPointerException: Object instance not found
at com.surrealdb.Value.getObject(Native Method)
at com.surrealdb.Value.getObject(Value.java:109)
at com.surrealdb.ValueClassConverter.convert(ValueClassConverter.java:175)
at com.surrealdb.Value.get(Value.java:212)
at dev.massi.ayase.database.model.AyaseUser.lambda$findById$0(AyaseUser.java:74)`

harry-xi:
> I noticed that `Response.take(Class<T>, int)` throws a `NullPointerException` if the query returns no results instead of returning null or an empty instance. This makes error handling more difficult in cases where no matching entry exists.
> 
> Would it be possible to handle this case more gracefully by returning `null` or an empty object instead of throwing an exception?

You should treat it as a separate issue as it is in fact relevant to the full range of scenarios involving Value to specific type conversions.

In fact I think in a scenario where None/Null might be returned in anticipation. A better way to handle this would be to get the original Value object directly from the Response then use its methods to determine if it is None/Null before converting it.

Further, the Value object provides comprehensive methods for determining whether the value in the Value is of a particular type, and they may be helpful.

massikhr:
> > I noticed that `Response.take(Class<T>, int)` throws a `NullPointerException` if the query returns no results instead of returning null or an empty instance. This makes error handling more difficult in cases where no matching entry exists.
> > Would it be possible to handle this case more gracefully by returning `null` or an empty object instead of throwing an exception?
> 
> You should treat it as a separate issue as it is in fact relevant to the full range of scenarios involving Value to specific type conversions.
> 
> In fact I think in a scenario where None/Null might be returned in anticipation. A better way to handle this would be to get the original Value object directly from the Response then use its methods to determine if it is None/Null before converting it.
> 
> Further, the Value object provides comprehensive methods for determining whether the value in the Value is of a particular type, and they may be helpful.

Thanks for your response! That makes sense. So instead of directly calling `take(Class<T>, int)`, I should first check the `Value` object returned from the `Response` to determine if it is `None` or `Null` and only then proceed with conversion?

Could you provide an example of how to properly check if the Value is None or Null before calling `get(Class<T>)`? That would be really helpful!

Appreciate the clarification! 🚀

src/main/rust/surreal.rs
```diff
@@ -207,6 +204,33 @@ pub extern "system" fn Java_com_surrealdb_Surreal_query<'local>(
...
+    params_keys: JObjectArray<'local>,
+    params_values: JLongArray<'local>,
+) -> jlong {
+    // Retrieve the Surreal instance
+    let surreal = get_surreal_instance!(&mut env, ptr, || 0);
+    // Retrieve the query
+    let query = get_rust_string!(&mut env,&query, || 0);
+    let keys = get_rust_string_array!(&mut env, params_keys, || 0);
+    let value_ptrs = get_long_array!(&mut env, &params_values, || 0);
+    let mut parems_map = BTreeMap::<String,&Value>::new();
```
emmanuel-keller, line 222:
```suggestion
    let mut params_map = BTreeMap::<String,&Value>::new();
```

src/main/rust/surreal.rs
```diff
@@ -207,6 +204,33 @@ pub extern "system" fn Java_com_surrealdb_Surreal_query<'local>(
...
+    for (key,value_ptr) in keys.into_iter().zip(value_ptrs) {
+        let value = get_value_mut_instance!(&mut env, value_ptr, || 0);
+        parems_map.insert(key, value);
```
emmanuel-keller, line 225:
```suggestion
        params_map.insert(key, value);
```

src/main/rust/surreal.rs
```diff
@@ -207,6 +204,33 @@ pub extern "system" fn Java_com_surrealdb_Surreal_query<'local>(
...
+    }
+    
+    let res = surrealdb_query::<&Value>(&surreal, &query, Some(parems_map));
```
emmanuel-keller, line 228:
```suggestion
    let res = surrealdb_query::<&Value>(&surreal, &query, Some(params_map));
```

emmanuel-keller (approved):
Great job! Just a typo!

harry-xi:
I found #90 also be resolve by this pr.

PR #139, opened 2026-04-15
Fix: surface selectLive() errors eagerly and improve LiveStream thread safety (#138)

### LINKED ISSUE #138: Bug: selectLive().next() can deadlock due to incorrect locking in Rust JNI layer
### Describe the bug

Calling LiveStream.next() after selectLive() can block forever (deadlock). The root cause is in the Rust JNI layer (live.rs), where the recv_mutex lock is held while blocking on receiving the next notification. If close() is called on another thread, or if the background stream thread interacts in certain timing, this can cause a deadlock because both sides try to acquire the same locks in opposite order. This deadlock can prevent closing the stream, drain JVM threads, or hang an application using surrealdb.java for live queries.

### Steps to reproduce

1. Start a live query using selectLive() in Java.
2. Run a while true loop, call next() and leave it blocked waiting for a notification.
3. Try updating a record and see that the notification never comes

### Expected behaviour

Calling next() should never block forever if close() is called, and closing a LiveStream should always cleanly awaken any waiting thread and terminate the blocking call.

### SurrealDB version

3.0.5

### SurrealDB Java SDK version

2.0.0

### Contact Details

_No response_

### Code of Conduct (repository /surrealdb.java)

emmanuel-keller:
Closes #138

## Investigation

The reported deadlock (lock-ordering issue in the Rust JNI layer) could not be confirmed — both `nextNative` and `releaseNative` acquire locks in the same order (`recv_mutex` then `rx_mux`), and the channel-close mechanism correctly unblocks `recv()`. Tests covering the exact reported scenarios (notifications after CREATE/UPDATE, `close()` unblocking `next()`, concurrent access) all pass.

However, the investigation uncovered two real issues that likely caused the reported symptoms.

### 1. Surface `selectLive()` subscription errors eagerly

`selectLive()` spawned a background thread and returned immediately, before the live query subscription was established. If the subscription failed (e.g. table does not exist), the error was silently queued in the channel and only surfaced on the first `next()` call — or worse, `next()` returned empty with no indication of what went wrong.

The background thread now signals readiness via a `std::sync::mpsc` channel. `selectLive()` blocks until the subscription is confirmed or an error is returned, so failures are thrown at call site.

### 2. Make `LiveStream.handle` volatile

`handle` was a plain `long` field read by `next()` and written by `close()` without synchronization. If `close()` ran on one thread (freeing native memory and zeroing `handle`), `next()` on another thread could read a stale non-zero value and pass a dangling pointer to `nextNative`.

## Code quality

- Documented the threading model, locking protocol, and shutdown sequence across `LiveStream.java`, `Surreal.java`, `lib.rs`, `live.rs`, and `surreal.rs`.
- Resolved all 45 clippy warnings in the Rust native layer (43 auto-fixes + 2 targeted suppressions).

## Tests

- `liveStreamReceivesCreateNotification` — blocks on `next()`, CREATEs a record, asserts notification arrives
- `liveStreamReceivesUpdateNotification` — same pattern for UPDATE
- `closeUnblocksBlockedNext` — `close()` from another thread unblocks a blocked `next()` without deadlocking
- `concurrentNextAndCloseDoesNotCrash` — stress test with multiple threads calling `next()` and `close()` concurrently
- `selectLiveOnNonExistentTableThrowsImmediately` — verifies the new eager error behavior

### DIFF
--- a/src/main/java/com/surrealdb/LiveStream.java
+++ b/src/main/java/com/surrealdb/LiveStream.java
@@ -13,7 +13,7 @@ public class LiveStream implements AutoCloseable {
 		Loader.loadNative();
 	}
 
-	private long handle;
+	private volatile long handle;
 
 	LiveStream(long handle) {
 		this.handle = handle;
--- a/src/main/java/com/surrealdb/LiveStream.java
+++ b/src/main/java/com/surrealdb/LiveStream.java
@@ -3,9 +3,27 @@ package com.surrealdb;
 import java.util.Optional;
 
 /**
- * Blocking iterator over live query notifications. Call {@link #next()} in a
- * loop and {@link #close()} when done. Implements {@link AutoCloseable} for
- * use in try-with-resources.
+ * Blocking iterator over live query notifications returned by
+ * {@link Surreal#selectLive(String)}.
+ *
+ * <p>Typical usage:
+ * <pre>{@code
+ * try (LiveStream stream = surreal.selectLive("person")) {
+ *     while (true) {
+ *         Optional<LiveNotification> n = stream.next();
+ *         if (!n.isPresent()) break;   // stream closed
+ *         process(n.get());
+ *     }
+ * }
+ * }</pre>
+ *
+ * <p><b>Thread safety:</b> {@link #next()} may be called from one thread while
+ * {@link #close()} is called from another. The {@code close()} call will
+ * unblock any thread currently waiting inside {@code next()}. The native
+ * handle is declared {@code volatile} so that the zeroing performed by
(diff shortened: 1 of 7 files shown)


joaoseidel:
Hey @emmanuel-keller, how are you?

I just tested the release with your fixes, but I figured out that the problem I was facing (which is still happening) does not seem to be related to deadlocks.

The issue appears to be that `selectLive` does not work over a WebSocket connection. I was only able to make it work with an in-memory connection.

Current behavior:
- `selectLive` works with an in-memory connection
- `selectLive` does not work over WebSocket
- The issue is still reproducible with the latest release

Could this be a limitation or a bug in the WebSocket connection implementation?

PR #157, opened 2026-05-21
Auto-load native library on first use of any Native-backed POJO

emmanuel-keller:
## Summary

- Move `Loader.loadNative()` into the `Native` base class static initializer so every JNI-backed POJO (`RecordId`, `Id`, `Array`, …) auto-loads the native library on first reference. Java's JLS §12.4.1 superclass-init guarantee makes this work for every current and future subclass.
- Remove the now-redundant `static { Loader.loadNative(); }` blocks from `Surreal` and `LiveStream`.
- Decouple `Loader` from `Surreal.class` by switching its resource lookup to `Native.class` (it no longer depends on the class it bootstraps).
- Add `PojoStandaloneTests` covering `new RecordId(...)`, `Id.from(...)`, and `Array.of(...)` without referencing `Surreal`/`LiveStream`.

## Why

Previously, the following code threw `UnsatisfiedLinkError` unless a `Surreal` instance had been created elsewhere in the JVM:

```java
import com.surrealdb.RecordId;

void main() {
    var recId = new RecordId("myRec", "foo"); // fails when no Surreal instance is running
    System.out.println("Record ID: " + recId);
}
```

Only `Surreal` and `LiveStream` ran `Loader.loadNative()` from a static initializer, so any POJO touched first failed. Users constructing record IDs for validation, tests, or builder code without a live `Surreal` connection were broken.

## Why this is safe

- `System.loadLibrary` is idempotent — duplicate calls in a process are no-ops.
- JLS §12.4.1 guarantees a superclass is initialized before its subclass, so `Native`'s initializer always runs before any subclass constructor or static native call.
- `Loader.loadNative()` itself is unchanged — only the *trigger point* moves earlier.
- `LiveStream` does not extend `Native`, but its constructor is package-private and the only public path to obtain one is via `Surreal.selectLive(...)`, which guarantees `Surreal` (and thus `Native`) is already loaded.

## Test plan

- [x] New `PojoStandaloneTests` (8 tests) all pass: covers `RecordId` (String, long, UUID keys), `Id.from(long|String|UUID|Object...)`, and `Array.of(...)`. Each test also calls `.toString()` and `.hashCode()` so the inherited `Native` methods are exercised.
- [x] The test class deliberately avoids importing `Surreal` / `LiveStream` and carries a header comment explaining the invariant (so a future contributor doesn't silently mask the regression by reaching for `Surreal`).
- [x] `./gradlew build` — full Java + record-test source sets pass. Only failing test is `ConnectionTests.connectWebsocket()`, which is preexisting and environmental (requires a SurrealDB server at `ws://localhost:8000`).
- [x] Spotless format check passes.

### DIFF
--- a/src/main/java/com/surrealdb/LiveStream.java
+++ b/src/main/java/com/surrealdb/LiveStream.java
@@ -31,10 +31,6 @@ import java.util.Optional;
  */
 public class LiveStream implements AutoCloseable {
 
-	static {
-		Loader.loadNative();
-	}
-
 	/**
 	 * Pointer to the native {@code LiveStreamChannel}. Zeroed by {@link #close()}
 	 * after the native resources have been released. Declared {@code volatile} so
--- a/src/main/java/com/surrealdb/Loader.java
+++ b/src/main/java/com/surrealdb/Loader.java
@@ -91,7 +91,7 @@ class Loader {
 
 	private static File extract(String path) throws IOException {
 		final String TargetsPath = "natives/" + path + "/" + SURREALDB_LIBNAME;
-		final URL Targets = Surreal.class.getClassLoader().getResource(TargetsPath);
+		final URL Targets = Native.class.getClassLoader().getResource(TargetsPath);
 		if (Targets == null) {
 			throw new RuntimeException("Couldn't find Targets: " + TargetsPath);
 		}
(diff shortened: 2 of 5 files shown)


emmanuel-keller:
@codex review

PR #166, opened 2026-05-29
Upgrade JNI (jni crate) 0.21.1 → 0.22.4

emmanuel-keller:
## What & why

Upgrades the native layer to the **`jni` Rust crate `0.22.4`**, a major API overhaul, in preparation for the 2.1.0 driver release. This is a Rust-side-only change — the **Java public API, JNI ABI / native method symbols, JDK 8 minimum, and the structured `ServerException` hierarchy are all unchanged**.

jni 0.22 raises the MSRV to **1.85**, which the repo's pinned toolchain satisfies.

## Key changes (jni 0.22 is a major overhaul)

- **`JNIEnv` split into `EnvUnowned` + `Env`.** Native methods now take `EnvUnowned` and run their body via `Env::with_env`. A small `with_env_body!` macro + `resolve_outcome` helper keep the existing raw `jobject`/`jstring`/`jlong` return types and the `SurrealError::exception` sink. (`EnvOutcome::resolve` requires `T: Default`, which raw pointers are not, so the outcome is resolved via `into_outcome()`.)
- **Signatures & strings:** method/constructor signatures → `jni_sig!`; class/method names → `jni_str!` (or `JNIString` for dynamic names); deprecated `Env::get_string` → `JString::try_to_string`.
- **`jni-sys 0.4` makes `jboolean = bool`** — `JValue::Bool` and the `jboolean` default closures/params were adjusted.
- **`JObject` → `JThrowable` is now the *fallible* `Env::cast_local`** (0.21's `.into()` was infallible). The structured-exception sink (`SurrealError::exception`) therefore falls back to a flat-string `throw_new` if **either** building the structured exception **or** the `JThrowable` cast fails — a server error can never be silently dropped. Typed `JObjectArray<JString>`; array APIs via `JObjectArray` / `JPrimitiveArray`.
- **Explicit `#[no_mangle]` export names are kept** (not `#[jni_mangle]`) so the 189 `Java_com_surrealdb_*` symbols stay byte-identical — zero `UnsatisfiedLinkError` risk. (Switching to `#[jni_mangle]` later is mechanical if preferred.)

## Dependency note

`Cargo.lock` carries **two `jni` versions**: `0.22.4` for this driver, and `0.21.1` pulled transitively by `rustls-platform-verifier` (Android cert verification) in surrealdb's rustls stack. The latter is not an orphan — nothing to prune.

## Drive-by fix

Corrected two copy-pasted `NullPointerException` labels in `Value.getDateTime` / `getDuration` (both previously read `"Geometry"`). Cosmetic and pre-existing — only affects the message on the type-mismatch error path.

## Verification

- `cargo build` / `cargo clippy --all-targets` / `cargo fmt --check` — clean, 0 warnings.
- `nm`: **189** `Java_com_surrealdb_*` export symbols, names byte-identical — no `UnsatisfiedLinkError` risk.
- Java test suite (in-memory + WebSocket server): **288 passed, 2 skipped** (OS-specific) — exercises the structured exception hierarchy and live queries over WebSocket through the real JNI boundary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Upgrade the native layer to the `jni` crate 0.22.4; the Java/JNI ABI, public API, and JDK 8 minimum are unchanged [#166](https://github.com/surrealdb/surrealdb.java/pull/166).
 - Upgrade to SurrealDB SDK 3.1.3 and bump the Rust toolchain to 1.95 [#165](https://github.com/surrealdb/surrealdb.java/pull/165).
 
 ## [2.0.3] - 2026-05-29
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2049,7 +2049,7 @@ dependencies = [
  "cesu8",
  "cfg-if",
  "combine",
- "jni-sys",
+ "jni-sys 0.3.0",
  "log",
  "thiserror 1.0.69",
  "walkdir",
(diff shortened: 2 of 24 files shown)


emmanuel-keller:
@codex review

PR #173, opened 2026-06-06
Add @SurrealName field mapping support

### LINKED ISSUE #171: Feature: Object conversion should support explicit field names via `@JsonProperty` or a naming strategy
### Summary

`ValueBuilder.convertObject` and `ValueClassConverter` currently appear to use raw Java field names when converting between Java objects and SurrealDB object values.

This makes it difficult to use idiomatic Java camelCase fields with snake_case database columns.

Example:

```java
public final class User {
	private String firstName;
	private Instant createdAt;
}
```

Currently this maps to:

```text
firstName
createdAt
```

but many schemas use:

```text
first_name
created_at
```

### Current behavior

The write path uses raw field names:

```java
String name = field.getName();
```

The read path similarly expects the database object keys to match raw Java field names.

This prevents common mappings such as:

```java
@JsonProperty("created_at")
private Instant createdAt;
```

### Expected behavior

The converter should support explicit field names and/or a configurable naming strategy.

At minimum, the following should work:

```java
public final class User {
	@JsonProperty("created_at")
	private Instant createdAt;
}
```

and map to the SurrealDB field:

```text
created_at
```

The same mapping should be honored in both directions:

```text
Java object -> SurrealDB object
SurrealDB object -> Java object
```

### Design considerations

There are a few possible approaches:

#### Option A: Support Jackson `@JsonProperty`

This is convenient for users already using Jackson annotations.

However, this introduces a dependency question:

- Should the SDK depend directly on Jackson annotations?
- Should Jackson support be optional?
- What happens when Jackson is not on the classpath?

#### Option B: Introduce a SurrealDB-specific annotation

Example:

```java
@SurrealName("created_at")
private Instant createdAt;
```

This avoids a Jackson dependency and gives the SDK full control over semantics.

#### Option C: Add a configurable naming strategy

Example:

```java
NamingStrategy.IDENTITY
NamingStrategy.SNAKE_CASE
```

This avoids field-level annotations for common naming conventions.

#### Option D: Support both explicit names and naming strategy

Recommended precedence:

```text
explicit annotation > configured naming strategy > raw field name
```

### Acceptance criteria

- Java field names can map to different SurrealDB object keys.
- The mapping works in both serialization and deserialization.
- Explicit field-level names take precedence over naming strategy.
- Raw `field.getName()` behavior remains the default for backward compatibility.
- Tests cover:
  - default field names,
  - explicit field names,
  - camelCase to snake_case mapping,
  - round-trip conversion.

- The implementation avoids introducing an unwanted hard dependency unless that is an intentional project decision.

### SurrealDB version

3.1.0

### SurrealDB Java SDK version

2.0.2

### Contact Details

_No response_

### Code of Conduct (repository /surrealdb.java)

PSmouz:
Add `@SurrealName` as a dependency-free way to map Java fields and record components to explicit SurrealDB object keys during conversion.

This lets users keep idiomatic Java camelCase names while storing snake_case or otherwise custom field names in SurrealDB:

```java
public final class User {
  @SurrealName("created_at")
  private ZonedDateTime createdAt;
}
```

The annotation is honored in both directions:

- Java object -> SurrealDB object
- SurrealDB object -> Java object

Unannotated fields keep the existing raw Java field-name behavior.

**Implementation Details**
Introduces `SurrealFieldNames`, an internal resolver for converter field metadata. The resolver centralizes:

- `@SurrealName` lookup
- raw field-name fallback
- duplicate resolved-name rejection
- blank annotation value rejection
- static/transient field exclusion
- inherited-field shadowing behavior

`ValueBuilder` now uses resolved field names during object serialization.

`ValueClassConverter` now builds deserialization lookups by resolved SurrealDB name instead of raw Java field name. Java field hiding is preserved: if a subclass field hides a superclass field, the superclass field is kept out of the lookup even when the subclass field is annotated to a different SurrealDB key.

Java record support is included by resolving annotations from record component backing fields while preserving the existing Java 8-compatible reflective record handling.

The write path keeps a raw-name fast path: classes with no declared `@SurrealName` fields continue using `field.getName()` directly and skip duplicate-name validation.

**Validation**
Adds POJO and record coverage for:

- annotated serialization
- annotated deserialization
- create/select round trips
- raw-name backward compatibility
- explicit-name precedence over raw keys
- unknown database key handling
- static/transient exclusion
- duplicate `@SurrealName` rejection
- blank `@SurrealName` rejection
- hidden superclass field regression
- record component mapping

**Verification**
Passed:

```text
./gradlew test recordTest spotlessCheck
```

(`recordTest` runs the record suites on JDK 16+; it is not part of the plain `test` task — see #175.)

Closes #171

**Behavior note**

Deserialization no longer assigns database keys to `static` or `transient` fields. Previously the read path matched any declared field by name, so a database key matching a `transient` (or even `static`) field would be reflectively set. Such fields are now excluded from the read lookup, mirroring the write path. Covered by `keepsStaticAndTransientFieldsUntouchedDuringDeserialization`.

### DIFF
--- a/src/main/java/com/surrealdb/SurrealFieldNames.java
+++ b/src/main/java/com/surrealdb/SurrealFieldNames.java
@@ -3,7 +3,9 @@ package com.surrealdb;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 final class SurrealFieldNames {
 
@@ -39,29 +41,28 @@ final class SurrealFieldNames {
 
 	static Map<String, Field> inheritedFieldsBySurrealName(final Class<?> clazz) {
 		final Map<String, Field> fields = new HashMap<>();
+		final Set<String> seenJavaNames = new HashSet<>();
 		Class<?> c = clazz;
 		while (c != null && c != java.lang.Object.class) {
 			for (final Field field : c.getDeclaredFields()) {
 				if (!isSerializableField(field)) {
 					continue;
 				}
+				// Java field hiding is based on the declared Java name, not the
+				// resolved SurrealDB name. Keep hidden superclass fields out even
+				// when the subclass field is annotated to a different key.
+				if (!seenJavaNames.add(field.getName())) {
+					continue;
+				}
 				final String name = nameFor(field);
-				// Preserve existing field-hiding behavior for unannotated raw names:
-				// the concrete class wins over a same-named superclass field.
-				if (fields.containsKey(name) && !isPlainRawName(field, name)) {
+				if (fields.containsKey(name)) {
 					throw duplicateName(name, fields.get(name), field);
 				}
-				if (!fields.containsKey(name)) {
-					fields.put(name, field);
-				}
+				fields.put(name, field);
 			}
 			c = c.getSuperclass();
 		}
 		return fields;
-	}
-
-	private static boolean isPlainRawName(final Field field, final String name) {
-		return field.getAnnotation(SurrealName.class) == null && field.getName().equals(name);
 	}
 
 	private static void putUnique(final Map<String, Field> fields, final String name, final Field field) {
--- a/src/main/java/com/surrealdb/SurrealFieldNames.java
+++ b/src/main/java/com/surrealdb/SurrealFieldNames.java
@@ -29,6 +29,15 @@ final class SurrealFieldNames {
 		return value;
 	}
 
+	static boolean hasDeclaredSurrealName(final Field[] fields) {
+		for (final Field field : fields) {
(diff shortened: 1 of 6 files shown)


emmanuel-keller:
@codex review

emmanuel-keller (approved):

emmanuel-keller:
Thanks @PSmouz for another high-quality contribution — and for raising the design discussion in #171 before diving in. The implementation landed exactly where we wanted it: dependency-free, backward compatible, and the edge-case handling (field hiding, duplicate/blank rejection, record component propagation) plus the breadth of test coverage made this a pleasure to review.

I pushed one small commit on top (948fb35): the new tests now use `query()` instead of the deprecated `queryBind()`, and there's an extra regression test pinning the read-side static/transient exclusion described in the behavior note. CI is green across the whole matrix — ready to go. 🎉

PR #178, opened 2026-06-10
Run the record test suite in CI

### LINKED ISSUE #175: Bug: CI never runs the recordTest source set
### Describe the bug

The `recordTest` source set (`src/recordTest/java`) is never executed in CI.

- `.github/workflows/test.yml` and `reports.yml` run `./gradlew -i test jacocoTestReport`
- `.github/workflows/cross.yml` runs `./gradlew -i integrationTest`
- The `recordTest` task is only wired into `check` (`build.gradle`: `tasks.named('check') { dependsOn 'recordTest' }`), and no workflow runs `check` or `recordTest`

Consequently `RecordCrudTests`, `RecordRoundTripTests`, and `SurrealNameRecordTests` (added in #173) only run on developer machines. They currently pass locally (verified on JDK 23 during the #173 review), but CI offers no protection against regressions in Java record support.

### Expected behaviour

CI should execute the record tests on the JDK versions that support records.

Suggested fix: in the `test.yml` matrix, run `./gradlew -i test recordTest jacocoTestReport` on the JDK 17/21/25 legs (keep plain `test` on 8/11 — `build.gradle` only registers the `recordTest` task graph on JDK 16+, so invoking it on older JDKs would fail).

### Notes

Pre-existing gap; surfaced while reviewing #173 because the new record-annotation tests land in this source set.

emmanuel-keller:
The `recordTest` source set (`RecordCrudTests`, `RecordRoundTripTests`, `SurrealNameRecordTests`) was only wired into `check`, while CI runs `gradlew test` — so the record tests never executed in CI.

**Changes**

- The test matrix now runs `./gradlew -i test recordTest jacocoTestReport` on every leg.
- `build.gradle` registers a **disabled placeholder** `recordTest` task on JDKs below 16, so the uniform invocation succeeds everywhere: on the 8/11 legs it reports `> Task :recordTest SKIPPED` instead of failing with *task not found*; on 17/21/25 the real record suites run.

**Verification**

- JDK 23 locally: `./gradlew recordTest` runs all record suites, green.
- Old-JDK path simulated by raising the version gate above the local JDK: `./gradlew test recordTest` → regular tests run, `recordTest` reports SKIPPED, build succeeds.
- This PR's own 8/11 matrix legs are the real-world proof of the placeholder path.

Closes #175

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,4 +80,4 @@ jobs:
       - name: Test
         run: |
           chmod +x gradlew
-          ./gradlew -i test jacocoTestReport
+          ./gradlew -i test recordTest jacocoTestReport
--- a/build.gradle
+++ b/build.gradle
@@ -242,6 +242,15 @@ if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_16)) {
     // Disable the auto-generated tasks for the recordTest source set on older JDKs.
     tasks.named('compileRecordTestJava') { enabled = false }
     tasks.matching { it.name == 'processRecordTestResources' }.configureEach { enabled = false }
+
+    // Disabled placeholder so `gradlew recordTest` is invocable on every JDK
+    // (the CI matrix runs it on all legs); it reports SKIPPED instead of
+    // failing with "task not found".
+    tasks.register('recordTest') {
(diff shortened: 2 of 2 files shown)


emmanuel-keller:
@codex review

PR #180, opened 2026-06-10
Cache per-class field metadata in the value converters

### LINKED ISSUE #176: Enhancement: cache per-class field metadata in the value converters
### Summary

Cache per-class field metadata used by the value converters instead of recomputing it on every conversion.

After #173:

- `SurrealFieldNames.inheritedFieldsBySurrealName(clazz)` builds a fresh `HashMap` (full hierarchy walk + annotation lookups + duplicate validation) for **every** deserialized object — including each element of a list and each nested object.
- On the write path, `SurrealFieldNames.hasDeclaredSurrealName` / `nameFor` re-scan annotations per serialized object.

This is the same cost order as the pre-#173 code (which did a per-key hierarchy walk using `NoSuchFieldException` as control flow), so it is not a regression — but bulk operations (large `SELECT` result sets, deeply nested graphs) would benefit from computing the metadata once per class.

### Suggested approach

- Use a `ClassValue<...>` keyed cache (avoids classloader leaks, unlike a static `ConcurrentHashMap<Class<?>, ...>`), storing per class:
  - the resolved-name → `Field` lookup for deserialization,
  - the "has `@SurrealName`" flag and resolved names for serialization,
  - validation results (blank/duplicate checks throw once, deterministically).
- No invalidation needed: annotations and declared fields are immutable at runtime.
- Record component metadata in `ValueClassConverter.convertRecord` (names/types/generic types arrays) could be cached the same way.

### Notes

Low priority; raised during the #173 review.

emmanuel-keller:
Field and record-component metadata is immutable at runtime, so resolve it once per class instead of on every conversion (#176). Previously `inheritedFieldsBySurrealName` rebuilt the hierarchy walk, annotation lookups, and duplicate validation for **every** converted object — including each element of a result list and each nested object.

**Changes**

- `SurrealFieldNames` caches the resolved-name field map in a `ClassValue` — tied to class lifetime, so classes stay unloadable (no static `Map<Class, ...>` leak). Fields are made accessible once at build time; the cached map is unmodifiable.
- **Failure semantics preserved**: if validation throws (blank/duplicate `@SurrealName`), `ClassValue` installs nothing and the same exception is rethrown on every subsequent attempt — covered by a new regression test.
- `ValueClassConverter` caches record metadata the same way: component names (with `@SurrealName` resolution), types, generic types, uniqueness validation, and the canonical constructor (`setAccessible` once).
- `ValueBuilder` now serializes through the cached map. The raw-name fast path from #173 is superseded: it existed to skip per-call annotation scans, which the cache eliminates for all shapes — after the first conversion of a class, the resolver is a single `ClassValue` lookup.

**Verification**

```text
./gradlew test recordTest spotlessCheck   # 322 tests; 1 env-only failure (connectWebsocket, needs local server)
```

Closes #176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Cache per-class field metadata in the value converters [#180](https://github.com/surrealdb/surrealdb.java/pull/180).
 - Support more Java datetime types in value conversion: `Instant`, `OffsetDateTime`, `LocalDateTime` (interpreted as UTC), `java.util.Date`, and the `java.sql` date types, in query bindings, `Array.of()`/`Id.from()`, and POJO/record round trips [#172](https://github.com/surrealdb/surrealdb.java/pull/172).
 - Add `@SurrealName` to map Java fields and record components to explicit SurrealDB object keys, honored in both serialization and deserialization [#173](https://github.com/surrealdb/surrealdb.java/pull/173).
 - Upgrade to SurrealDB SDK 3.1.4 [#177](https://github.com/surrealdb/surrealdb.java/pull/177).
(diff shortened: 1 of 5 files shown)


emmanuel-keller:
@codex review

PR #183, opened 2026-06-19
Add full geometry-type support (read + write)

emmanuel-keller:
## Summary

Resolves #99. The README's "Planned Features" listed *"All Geometry types (actually only points)"* — this PR delivers the feature instead of just rewording it.

Extends `Geometry` from Point-only / read-only to **all seven geometry types, read + write**: Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, and GeometryCollection.

## Public API (`Geometry`)

- **Type discrimination:** `getType()` → `"Point" | "LineString" | … | "GeometryCollection"`, plus `isLineString()` / `isPolygon()` / `isMultiPoint()` / `isMultiLineString()` / `isMultiPolygon()` / `isGeometryCollection()`. Existing `isPoint()` / `getPoint()` are unchanged.
- **Readers:** `getLineString()`, `getMultiPoint()` → `List<Point2D.Double>`; `getPolygon()`, `getMultiLineString()` → `List<List<Point2D.Double>>` (polygon ring 0 = exterior, the rest are holes); `getMultiPolygon()` → `List<List<List<Point2D.Double>>>`; `getGeometryCollection()` → `List<Geometry>`.
- **Factories:** `point`, `lineString`, `polygon`, `multiPoint`, `multiLineString`, `multiPolygon`, `geometryCollection`. Geometries now serialize through `create` / `update` content and bound params (wired via `ValueMut` / `ValueBuilder` / `ValueBoxing`).

Coordinates use `java.awt.geom.Point2D.Double`, where `x` is the longitude and `y` is the latitude.

## Implementation notes

- **Reads** reuse the SDK's `Geometry::as_coordinates()` and the existing `Value` / `Array` traversal (coordinates decode as `getDouble()`), keeping the new Rust surface small. `GeometryCollection` returns child `Geometry` pointers so each child keeps its own type.
- **Writes** build `geo` types directly via `Geometry::from_*`. Adds `geo = "0.32"` — the exact version `surrealdb-types` uses, so the `geo::*` types unify with `surrealdb::types::Geometry`.

## Bug fix

`Geometry` declared a `deleteInstance` native with **no matching Rust symbol** — finalizing any `Geometry` would have thrown `UnsatisfiedLinkError` and leaked the `Arc<Value>`. Added `Java_com_surrealdb_Geometry_deleteInstance`.

## Tests

[...]

PR #188, opened 2026-07-10
Support JDBC-style connection URLs in connect(String)

### LINKED ISSUE #105: Feature: support JDBC url style
### Is your feature request related to a problem?

most java applications use the jdbc url style to config the connection to a database

### Describe the solution

support the jdbc url style, please

### Alternative methods

nothing else

### SurrealDB version

SurrealDB 2.1.4 macOS  arrch64

### Contact Details

dashujuxuexi@163.com

### Code of Conduct

welpie21:
Closes #105

## What this adds

`Surreal.connect(String)` now accepts a JDBC-style connection URL of the form
`jdbc:surrealdb:...` as a convenience shorthand, in addition to the native
connect strings it already understood (`ws://...`, `memory`, `surrealkv://...`,
etc). When a JDBC-style URL is given, `connect()` transparently connects,
signs in (if credentials were supplied), and selects namespace/database (if
supplied) - equivalent to chaining `connect()` + `signin(Credential)` +
`useNs()` + `useDb()` manually.

Example URLs:

- Remote engine over WebSocket, with embedded user/pass/ns/db:
  `jdbc:surrealdb://root:root@localhost:8000/my_ns/my_db`
- In-memory embedded engine, with ns/db passed as query params:
  `jdbc:surrealdb:memory?ns=my_ns&db=my_db`
- Embedded SurrealKV (file-backed) engine:
  `jdbc:surrealdb:surrealkv:///var/lib/surreal/data?ns=my_ns&db=my_db`

`http://`/`https://` remote schemes and query-string aliases
(`user`/`username`, `pass`/`password`, `ns`/`namespace`, `db`/`database`) are
also supported; see the Javadoc on `Surreal.connect(String)` and the new
`JdbcConnectionUrl` class for the full grammar.

**Credentials supplied this way always authenticate as root.** Namespace- or
database-scoped users, or record users, cannot be represented in this
shorthand - callers that need those should keep using the existing
`connect(String)` + `signin(Credential)` path directly.

## Tests added

- `src/test/java/com/surrealdb/JdbcConnectionUrlTests.java` (new, 29 unit
  tests for the parser): remote ws/wss/http/https URLs with and without
  userinfo/path/port, the default-to-`ws` rule for a bare `jdbc:surrealdb://`,
  query-string credential/ns/db aliases and their precedence vs userinfo/path,
  `memory`/`mem://` handling, arbitrary embedded/file engine passthrough
  (`surrealkv://`, `rocksdb://`), and error cases (missing prefix, wrong
  subprotocol, null input, missing host, too many path segments, empty
  namespace/database path segments, IPv6 host literals, percent-encoded
  passwords containing reserved delimiter characters, and scheme
  case-normalization).
- `src/test/java/com/surrealdb/ConnectionTests.java` (modified): added
  `jdbcConnectWebsocket`, `jdbcConnectWebsocketUppercaseScheme`,
  `jdbcConnectMemory`, and `jdbcConnectSurrealKV`, exercising the JDBC URL
  path end-to-end against a real SurrealDB instance.

No native/Rust code was touched - `connect()` on the native side is an
unmodified thin passthrough to the `surrealdb::engine::any` connector; all of
the new logic lives in the Java-side URL parser.

## Why draft

Opening this as a **draft** for maintainer review: this repo's
`CONTRIBUTING.md` asks that new feature ideas normally go through a GitHub
discussion first before a pull request. This PR was scoped directly from an
open, unclaimed feature-request issue (#105) rather than a prior discussion
thread, so flagging that here for maintainers to weigh in on scope/design
before this is considered for merge.


src/main/java/com/surrealdb/JdbcConnectionUrl.java
```diff
@@ -0,0 +1,348 @@
...
+		}
+		final QueryParams queryParams = QueryParams.parse(query);
+		return new JdbcConnectionUrl(remainder, queryParams.ns, queryParams.db, queryParams.user, queryParams.pass);
+	}
+
+	private static String firstNonNull(String preferred, String fallback) {
+		return preferred != null ? preferred : fallback;
+	}
+
+	private static String decode(String value) {
```
emmanuel-keller, line 338: **Blocking.** `URLDecoder` performs `application/x-www-form-urlencoded` decoding, so it turns a literal `+` into a space. That's correct for query-string values but wrong for URI userinfo and path components.

Concrete failure: `jdbc:surrealdb://root:pa+ss@localhost:8000` parses the password as `pa ss`, so the root sign-in fails auth with no diagnostic. `+` is a common password character, and the existing `percentEncodedPasswordWithSpecialCharacters` test covers `@ / :` but not `+`, so CI won't catch it.

Suggest decoding userinfo/path with a percent-only decoder (expand `%XX`, leave `+` untouched) and keeping form-decoding for the query-string values.

src/main/java/com/surrealdb/Surreal.java
```diff
@@ -172,12 +172,50 @@ final boolean equals(long ptr1, long ptr2) {
...
 	public Surreal connect(String connect) {
+		if (connect != null && connect.length() >= 5 && connect.substring(0, 5).equalsIgnoreCase("jdbc:")) {
+			final JdbcConnectionUrl parsed = JdbcConnectionUrl.parse(connect);
+			connect(getPtr(), parsed.getNativeConnectString());
+			namespace = null;
+			database = null;
+			if (parsed.getUsername() != null) {
+				signin(new RootCredential(parsed.getUsername(), parsed.getPassword()));
+			}
+			if (parsed.getNamespace() != null) {
```
emmanuel-keller, line 215: Non-blocking: `useDb(...)` is nested inside the `getNamespace() != null` guard, so a database provided without a namespace (reachable via `?db=foo`) is silently dropped. SurrealDB can't select a db without a ns anyway, but silently ignoring the param is surprising — consider throwing or documenting.

src/main/java/com/surrealdb/JdbcConnectionUrl.java
```diff
@@ -0,0 +1,348 @@
...
+		String userinfoUser = null;
+		String userinfoPassword = null;
+		final int atIdx = remainder.lastIndexOf('@');
+		String hostPort = remainder;
+		if (atIdx >= 0) {
+			final String userinfo = remainder.substring(0, atIdx);
+			hostPort = remainder.substring(atIdx + 1);
+			final int colonIdx = userinfo.indexOf(':');
+			if (colonIdx >= 0) {
+				userinfoUser = decode(userinfo.substring(0, colonIdx));
```
emmanuel-keller: Non-blocking: empty userinfo (`jdbc:surrealdb://@localhost:8000`) yields `username=""` rather than null, which then triggers a root sign-in with empty credentials. Consider normalizing empty userinfo to null (and/or requiring a password when a username is present).

src/test/java/com/surrealdb/JdbcConnectionUrlTests.java
```diff
@@ -0,0 +1,241 @@
...
+		// with a single trailing slash ("/ns"), which legitimately means "namespace
+		// only, no database" (see trailingSlashAfterNamespaceOnlyIsAllowed).
+		assertThrows(IllegalArgumentException.class,
+				() -> JdbcConnectionUrl.parse("jdbc:surrealdb://localhost:8000/ns//"));
+	}
+
+	@Test
+	void trailingSlashAfterNamespaceOnlyIsAllowed() {
+		// A single trailing slash with no database segment is fine (not the same as
+		// an empty database segment): "ns/" means "namespace=ns, no database".
```
emmanuel-keller, line 219: Non-blocking: `trailingSlashAfterNamespaceOnlyIsAllowed` actually parses `.../ns` (no trailing slash); the real `.../ns/` case isn't exercised. Worth renaming or adding the actual trailing-slash case.

emmanuel-keller (changes requested):
**Review — request changes.**

Nice addition overall — the parser is well structured and comes with strong coverage (29 parser unit tests plus integration tests for ws/memory/surrealkv and the uppercase-scheme regression). There's one blocking issue (inline, on the `URLDecoder` usage) plus a few smaller points.

**Also:** no `CHANGELOG.md` entry was added — recent PRs (#183/#184/#185) each add one referencing the PR number; please add one here too.

src/main/java/com/surrealdb/JdbcConnectionUrl.java
welpie21: Fixed in 187c7ec. Reproduced first: `jdbc:surrealdb://ro+ot:pa+ss@localhost:8000` parsed the password as `pa ss` (and `/my+ns/my+db` path segments were mangled the same way). Added a private `percentDecode()` that expands `%XX` (incl. multi-byte UTF-8, by re-encoding literal `+` as `%2B` before delegating to `URLDecoder`) and used it for userinfo and path segments; query-string values keep form-decoding, where `+` means space. New tests cover `+` in userinfo and path, `%2B`, multi-byte UTF-8 in userinfo/path, and pin the `+`-to-space behavior for query values.

src/main/java/com/surrealdb/Surreal.java
welpie21: Fixed in 187c7ec — went with throwing. `JdbcConnectionUrl.parse()` now validates cross-field consistency after merging the userinfo/path and query-string sources, so `?db=foo` with no namespace from either source throws `IllegalArgumentException("database supplied without a namespace")` instead of being silently dropped; this matches the parser's existing fail-loud handling of empty segments and missing hosts, and covers the memory/embedded forms too. Mixed sources remain valid (`/ns?db=foo`). With the invariant enforced at parse time, the `useDb` call in `connect()` is un-nested from the namespace guard. Tests: `databaseWithoutNamespaceThrows`, `databaseWithoutNamespaceThrowsForMemory`, `pathNamespaceWithQueryStringDatabase`.

src/main/java/com/surrealdb/JdbcConnectionUrl.java
welpie21: Fixed in 187c7ec. An empty userinfo (`@host`, and `:@host`) is now normalized to null credentials, so no sign-in is attempted, and it no longer shadows query-string credentials (`jdbc:surrealdb://@host?user=root&pass=root` works). Also took the second suggestion: credentials must be complete after merging both sources — a username without a password now throws (passing a null password through to the native `signin` fails with an unhelpful null-pointer error, so failing fast at parse seemed better), as does a password without a username; one source may still complete the other (`root@host?pass=secret`). An explicitly empty password (`root:@host`) is still accepted and fails at the server with a genuine auth error. Semantics are documented in the class javadoc and `connect()`'s `@throws`.

src/test/java/com/surrealdb/JdbcConnectionUrlTests.java
welpie21: Fixed in 187c7ec. `trailingSlashAfterNamespaceOnlyIsAllowed` now parses the actual `.../ns/` form (it does parse as namespace-only, so this was a coverage gap, not a code bug — the bare `.../ns` case was already covered by `wsWithNamespaceOnlyPath`), and added `trailingSlashAfterNamespaceAndDatabaseIsAllowed` for `.../ns/db/`.

welpie21:
Added the missing `CHANGELOG.md` entry under `## [Unreleased]`, referencing this PR: 16de858.

PR #190, opened 2026-07-10
Add auth() and close remaining auth-parity gaps with the JS SDK

welpie21:
## Summary

This branch started as a single fix (`auth()`) and grew into a full pass over authentication parity between the JS SDK (golden reference) and the Java SDK, comparing both public and internal machinery. It now contains four changes:

1. `Surreal.auth()` / `Surreal.auth(Class<T>)` — fetch the authenticated record user (original scope).
2. `AccessKeyCredential` — sign in with a `TYPE BEARER` access grant key.
3. `AccessCredential` — sign in via a named system-level `ACCESS` method with username/password.
4. `Surreal.authenticate(Token)` — a real refresh-token rotation, not just a re-attach.

A fourth candidate considered during the research pass, a public `accessToken`/`refreshToken` getter on `Surreal`, was rejected: the Rust SDK has no API to introspect the currently-active token after signin/signup/authenticate, and the JS SDK's own `revoke()` RPC method is similarly unreachable from its public surface — neither is a gap worth porting.

## 1. `auth()` — fetch the authenticated record user

**The gap.** The Java SDK had no way to fetch the currently-authenticated record user. The JS SDK (golden reference) exposes this as `auth()`; SurrealDB's underlying RPC method is named `info` (the v1 JS SDK also called its method `info()`, before it was renamed to `auth()`), implemented as `SELECT * FROM ONLY $auth`. Java users had to hand-write the `$auth` query. The Java method is named `auth()`, matching the current JS SDK method name rather than the RPC method name or the v1 JS name.

**The fix.** `Surreal.auth()` returning `Optional<Value>` and `Surreal.auth(Class<T>)` returning `Optional<T>`, following the `select(RecordId)` / `select(Class, RecordId)` conventions. Implemented as a native JNI method (`Java_com_surrealdb_Surreal_auth`) mirroring `selectRecordId`: it runs `SELECT * FROM $auth` (deliberately without `ONLY`, so an unauthenticated session yields an empty array mapped to `Optional.empty()` instead of the `ONLY`-keyword error) and returns the single record when present.

**Tests.** `AuthTests` (memory engine): signup then `auth()`/`auth(Class)` return the record; both are empty when unauthenticated, after `invalidate()`, and when the record exists but table permissions hide it from the authenticated user.

## 2. `AccessKeyCredential` — signin with a `TYPE BEARER` access grant key

**The gap.** SurrealDB has a first-class auth flow: `DEFINE ACCESS x ON DATABASE TYPE BEARER FOR USER ...` + `ACCESS x ON DATABASE GRANT FOR USER someuser` issues a key like `surreal-bearer-<id>-<random>`. Signing in with that key sends RPC `signin` with `{ns?, db?, ac: "x", key: "<key>"}` — verified directly against a live SurrealDB 3.2.1 server. The JS SDK supports this as `AccessBearerAuth = {namespace?, database?, access, key}`, wire-mapped 1:1 in `internal/build-rpc-auth.ts` — notably this branch does **not** fall back to session ns/db like record auth does; ns/db are sent exactly as given (possibly absent).

Java had no equivalent. There **is** a class called `com.surrealdb.signin.BearerCredential`, but it does something completely different and is misleadingly named: it just calls the plain `authenticate(existingToken)` RPC to attach an already-issued JWT to the session — i.e. it's really "reauth with an existing token," unrelated to SurrealDB's `TYPE BEARER` access-grant flow.

**The fix.** A new class, `AccessKeyCredential` (fields: nullable `namespace`/`database`, non-null `access`, non-null `key`), implementing `Credential`, wired into `Surreal.signin(Credential)`. `BearerCredential` is untouched (additive only) but its Javadoc, and `AccessKeyCredential`'s, now call out the naming collision explicitly in both directions.

At the native layer, the vendored `surrealdb` 3.1.5 crate's built-in `Credentials<Signin>` impls are only `Root`/`Namespace`/`Database`/`Record<P>` — none support a bearer key. `Credentials<Action>` is a public marker trait (`SurrealValue` bound only) and `Surreal::signin` is generic, so a small local Rust struct (`AccessKeyCredentials` in `surreal.rs`) implements `SurrealValue` manually (mirroring `Record<P>`'s manual impl: build an `Object`, insert fields, wrap as `Value::Object`) and the empty marker `Credentials<Signin>` — allowed under Rust's orphan rules since the struct is local. Only `into_value` needs a real implementation; `from_value` is never invoked on the signin path (`Signin::into_future` only calls `.into_value()`/`.into_object()`) so it returns an error.

Namespace/database are genuinely optional on the wire (no session fallback, matching the JS SDK): the JNI function takes nullable `JString` parameters via a new `get_optional_rust_string` helper (`JString::is_null()` → `None`, otherwise delegate to `try_to_string`), since the existing `get_rust_string!` macro assumes non-null.

**Tests.** `AccessAuthTests`: a database-level bearer grant signin (asserting the resulting session is authenticated via `$access`), a namespace-level bearer grant with the database omitted, an invalid-key case throwing `SurrealException`, and a case proving there is no session fallback (namespace/database are set via `useNs`/`useDb`, but the credential still fails when it omits them — because they are never sent).

## 3. `AccessCredential` — signin via a named system-level `ACCESS` method

**The gap.** The JS SDK's `AccessSystemAuth = {namespace?, database?, username, password, access}` is structurally identical to `RootAuth`/`NamespaceAuth`/`DatabaseAuth` (`{user, pass}` wire shape) plus a mandatory `access` field (→ wire `ac`): a named, user-defined system-level `ACCESS` method verified with username/password, distinct from the built-in root/namespace/database user tables. `build-rpc-auth.ts` only throws `MissingNamespaceDatabaseError` if `database` is given without `namespace`; it never falls back to session state otherwise.

Java had no equivalent — `RootCredential`/`NamespaceCredential`/`DatabaseCredential` have no `access` field at all.

**The fix.** A new class, `AccessCredential` (fields: nullable `namespace`/`database`, non-null `username`/`password`/`access`), implementing `Credential`, with three constructors of growing arity (no ns/db, namespace-only, namespace+database) mirroring the fixed `username, password, ...` parameter order already used by `RootCredential`→`NamespaceCredential`→`DatabaseCredential`. Wired into `Surreal.signin(Credential)`. The ns-without-db validation is a plain Java check before the native call, mirroring how `RecordCredential`'s ns/db resolution is structured in `Surreal.signin`/`signup`, and throws `SurrealException` with a message naming the requirement (matching the JS `MissingNamespaceDatabaseError` behavior).

At the native layer this is another small local struct (`AccessCredentials`), built the same way as `AccessKeyCredentials` — nullable ns/db, plus `user`/`pass`/`ac` always present.

**Tests.** `AccessAuthTests`: a `TYPE RECORD` access method verified against `$user`/`$pass` (SurrealDB's core dispatch for a named `ACCESS` method combined with username/password credentials routes through the same `Record`-access mechanism regardless of whether the field names are `email`/`password` or `user`/`pass` — confirmed by reading `surrealdb-core`'s `iam::signin::db_access`, which only recognizes `Record` or `Bearer` access types on this path), asserting the resulting session is authenticated via `$access`; wrong credentials throwing; and a database-without-namespace case throwing before any request is made.

## 4. `Surreal.authenticate(Token)` — real refresh-token rotation

**The gap.** The JS SDK's `authenticate(token: Token | Tokens)` branches: given a `{access, refresh}` pair with a truthy `refresh`, it performs an actual RPC-level refresh that **rotates** both tokens (new access *and* new refresh returned by the server) — gated by SurrealDB ≥ 3.0.0. Given just a plain access string (or an object without a refresh), it does the plain `authenticate` RPC call (re-validate/attach a token, no rotation) and returns `{access}` as an echo (not server-confirmed).

Java's `com.surrealdb.signin.Token` class already models `access` + nullable `refresh` — clearly meant to support a refresh flow, but nothing wired it up. `Surreal.authenticate(String token)` only ever did the plain path.

**The fix.** A new overload, `Surreal.authenticate(Token)` (the existing `authenticate(String)` is unchanged — this is additive). When `token.getRefresh()` is non-null, a new native JNI function (`authenticateRefresh`) builds `surrealdb::opt::auth::Token{access, refresh: Some(...)}` and calls `surreal.authenticate(token).refresh().await` — the vendored SDK's own two-mode builder (`Authenticate<C>::refresh()` switches from `Command::Authenticate` to `Command::Refresh`) — returning the new, rotated `Token` via the existing `new_token_object` helper. Otherwise it calls the *existing* native `authenticate` function with `token.getAccess()` and returns `new Token(token.getAccess(), null)` — an echo, matching the JS SDK's `{access}` return, not fabricated. No client-side SurrealDB-version gate is added; if the server doesn't support refresh, the native call surfaces a `SurrealException` naturally, consistent with how this SDK handles other server-version-dependent behavior elsewhere in the authentication surface.

One implementation note: the vendored SDK's `Authenticate<C, RefreshToken>` guards against calling `.refresh()` on a token with no refresh component, returning `Error::internal("Missing refresh token")`. To keep that defensive path both reachable and meaningfully testable from Java (`Surreal.authenticate(Token)` only ever calls the refresh-native-function when `Token.getRefresh()` is already non-null), the native function treats an *empty* refresh string as "no refresh token" the same way — so a `Token` with a non-null but empty refresh cleanly triggers that guard and surfaces as a `SurrealException` rather than a crash, without changing normal-path behavior (real refresh tokens are never empty strings).

**Tests.** `AccessAuthTests`: signup against a `TYPE RECORD ... WITH REFRESH` access method, call `authenticate(token)`, and assert the returned access **and** refresh are both different from the input (proving rotation, not an echo) — mirrors the JS integration test at `packages/tests/surrealdb/integration/authentication.test.ts:242-261`. Also the plain-string-equivalent path (`authenticate(new Token(access, null))`) returning `{access, null}` without rotation, and calling the refresh path with an empty refresh surfacing a clean `SurrealException` rather than crashing.

## Out of scope (considered, not implemented)

- A public `refreshToken`/`accessToken` getter on `Surreal` (see Summary).
- `reset()` (JS's combined auth+ns/db+vars clear) — session-state scope, not strictly authentication; `invalidate()` + `useDefaults()` + `unset()` already cover the individual pieces.
- Renaming, removing, or repurposing `BearerCredential` — it is existing, released public API; this branch only adds a Javadoc cross-reference to it.

## Full test suite

348 tests, 0 failures — `AuthTests` (5 tests) plus `AccessAuthTests` (10 tests) account for 15 of those; `spotlessCheck` and `cargo fmt --all --check` are both clean.


src/main/java/com/surrealdb/Surreal.java
```diff
@@ -420,6 +422,51 @@ public Surreal invalidate() {
...
+	 * it (e.g.
+	 * {@code DEFINE TABLE user PERMISSIONS FOR select WHERE id = $auth.id}).
+	 * <p>
+	 * For more details, check the <a href=
+	 * "https://surrealdb.com/docs/surrealdb/security/authentication#record-users">authentication
+	 * documentation</a>.
+	 *
+	 * @return an Optional containing the authenticated record user's Value, or an
+	 *         empty Optional when the session is not authenticated as a record user
+	 *         (e.g. unauthenticated, invalidated, or a system user) or the record
```
emmanuel-keller, line 511: The Javadoc says empty is also returned when "the record is not visible to it" (permissions). That's plausible but untested and server-version-dependent — the only permissions case covered is the permissive `WHERE id = $auth.id`. Consider softening the wording or adding a test.

emmanuel-keller (commented):

Clean and faithful to the existing `select(RecordId)` pattern on both sides. I checked the failure modes: `get_surreal_ref!` (not the cloning variant) is used so `$auth` resolves on the real session; unauthenticated / system-user (root/ns/db) sessions yield `[]` → `Optional.empty()` on both sides; there's no `unwrap`/`expect` and all error paths route through the exception sink; the returned `Value` is owned/cleaned up exactly like `select`.

**Two points:**
1. Version: `build.gradle` is still `2.1.2`, which is already released (commit bc3812c). This net-new public API needs a minor bump (e.g. `2.2.0-SNAPSHOT`) — but that's really a batch decision across the currently-open PRs, not just this one.
2. Test gap: no explicit system-user (root/ns/db) case asserting empty — low risk since `info_unauthenticated_returnsEmpty` hits the same `$auth`=NONE path, but worth adding (use a ns/db user, since the memory engine may reject root sign-in).

src/main/java/com/surrealdb/Surreal.java
welpie21: Added a test rather than softening the wording: `info_recordNotVisibleToUser_returnsEmpty` (8e375e8) signs up a record user against a table defined with `PERMISSIONS FOR select NONE`, asserts the session is still authenticated (`RETURN $auth` yields a `user` record id), and then asserts both `info()` and `info(Class)` return an empty Optional. So the behavior matches the Javadoc on the embedded engine (SurrealDB 3.1.5): `SELECT * FROM $auth` returns an empty result set when permissions hide the record, which `info()` maps to `Optional.empty()`. Kept the Javadoc as-is now that the claim is pinned by a test.

